PR #2, opened 2019-03-13
Define basic interfaces and data structures

michaelwoerister:
This PR defines the basic data types and interfaces of the library.
`tests/end_to_end.rs` contains an example of how I roughly imagine the public API of the crate to look like.

~~This is not meant to be merged yet (it doesn't even compile).~~

cc @wesleywiser

### DIFF
--- a/src/mmap_serialization_sink.rs
+++ b/src/mmap_serialization_sink.rs
@@ -1,0 +1,18 @@
+use crate::serialization::{Addr, SerializationSink};
+use std::path::Path;
+
+pub struct MmapSerializationSink {}
+
+impl SerializationSink for MmapSerializationSink {
+    fn from_path(_path: &Path) -> Self {
+        unimplemented!()
+    }
+
+    #[inline]
+    fn write_atomic<W>(&self, _num_bytes: usize, _write: W) -> Addr
+    where
+        W: FnOnce(&mut [u8]),
+    {
+        unimplemented!()
+    }
+}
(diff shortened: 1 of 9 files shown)


src/mmap_serialization_sink.rs
```diff
@@ -0,0 +1,18 @@
+use crate::serialization::{Addr, SerializationSink};
+use std::path::Path;
+
+pub struct MmapSerializationSink {}
```
wesleywiser: It might be worth implementing a plain sink on top of regular files first and then seeing what the performance difference is from using `mmap`'d files. Just an idea though.

src/profiler.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::serialization::SerializationSink;
+use crate::stringtable::{SerializableString, StringId, StringTableBuilder};
+use std::path::Path;
+use std::sync::Arc;
+use std::time::Instant;
+
+pub struct Profiler<S: SerializationSink> {
+    event_sink: Arc<S>,
+    string_table: StringTableBuilder<S>,
+    start_time: Instant,
```
wesleywiser: We might want to change this to be a pair of `SystemTime` and `Instant` values like we did in `rustc`. That way we have an "absolute" start time and combine data from several different profiles into a single event stream.

src/profiling_data.rs
```diff
@@ -0,0 +1,17 @@
...
+use std::path::Path;
+
+pub struct ProfilingData {}
+
+impl ProfilingData {
+    pub fn new(_path_stem: &Path) -> ProfilingData {
+        unimplemented!()
+    }
+
+    pub fn iter_events<'a, F>(&'a self, mut _f: F)
```
wesleywiser: I'd like to see this return an iterator but that's not really important right now.

PR #5, opened 2019-03-25
Implement missing parts of api & get example working

wesleywiser:
- Resolves some `unimplemented!()` calls
- Adds a simple file based sink
- Adds an event `Iterator`
- Moves the repo into a Cargo Workspace configuration

cc @michaelwoerister


src/profiling_data.rs
```diff
@@ -1,17 +1,73 @@
...
+        if raw_idx_end > self.data.event_data.len() {
+            return None;
+        }
+
+        self.curr_event_idx += 1;
+
+        let raw_event_bytes = &self.data.event_data[raw_idx..raw_idx_end];
+
+        let raw_event = unsafe {
+            &*(raw_event_bytes.as_ptr()  as *const RawEvent)
```
michaelwoerister: We can't do this because the data at `raw_event_bytes` might not have the right alignment. In order to make this safe we have to allocate an event on the stack (e.g. `let mut event = RawEvent::default();`) and then copy the bytes from `raw_event_bytes` into the stack allocated `RawEvent`, i.e. the reverse of this:
https://github.com/rust-lang/measureme/blob/54c175dae6048433106d6099e446c6fd4bb898cd/src/profiler.rs#L59-L73

The important thing is that the `copy_from_slice` calls is done with `u8` slices (which are always safe to work with because of their alignment of 1) instead of pointers to `RawEvent`.

src/stringtable.rs
```diff
@@ -223,14 +223,14 @@ impl<'st> StringRef<'st> {
...
     // TODO: Replace with something lazy
-    string_data: &'data [u8],
+    string_data: Vec<u8>,
     index: FxHashMap<StringId, Addr>,
 }
 
-impl<'data> StringTable<'data> {
-    pub fn new(string_data: &'data [u8], index_data: &[u8]) -> StringTable<'data> {
+impl<'data> StringTable {
+    pub fn new(string_data: Vec<u8>, index_data: Vec<u8>) -> StringTable {
```
michaelwoerister: `index_data` can stay a slice since we only construct the `index` table from it but don't hold on to it.

measureme/src/profiling_data.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        let raw_event_bytes = &self.data.event_data[raw_idx..raw_idx_end];
+
+        let raw_event = unsafe {
+            &*(raw_event_bytes.as_ptr()  as *const RawEvent)
+        };
+
+        let string_table = &self.data.string_table;
+
+        let mut timestamp = SystemTime::UNIX_EPOCH;
```
michaelwoerister: The way `Profiler::record_event()` works at the moment, the timestamp contains the number of nanoseconds since `Profiler::start_time`, not since `SystemTime::UNIX_EPOCH`.

src/profiling_data.rs
```diff
@@ -1,17 +1,73 @@
...
+            event_data,
+        }
     }
 
-    pub fn iter_events<'a, F>(&'a self, mut _f: F)
-    where
-        F: FnMut(&Event<'a>),
-    {
-        unimplemented!()
+    pub fn iter(&self) -> impl Iterator<Item = Event<'_>> {
```
michaelwoerister: One advantage of the callback based approach is that memory allocations for strings and stuff can be re-used, which is not possible with an iterator (I think).

measureme/src/profiling_data.rs
```diff
@@ -0,0 +1,73 @@
...
+        let raw_idx_end = raw_idx + mem::size_of::<RawEvent>();
+        if raw_idx_end > self.data.event_data.len() {
+            return None;
+        }
+
+        self.curr_event_idx += 1;
+
+        let raw_event_bytes = &self.data.event_data[raw_idx..raw_idx_end];
+
+        let raw_event = unsafe {
```
michaelwoerister: Whoops, the old comment seems to be overwritten. See here: https://github.com/rust-lang/measureme/pull/5/commits/392d882aca536c4ce8502fdddbf25cb9de680355#r268630278

michaelwoerister:
Thanks, @wesleywiser! Looks good except for the pointer alignment issue. 

I'm not sure about using an `Iterator` for iterating over the events in `ProfilingData`. It makes lifetimes a bit more complicated. Why do you prefer it over the callback approach?

measureme/src/profiling_data.rs
wesleywiser: I think the reason I changed this was because there was no way to reconstruct the `Instant` values since the only `Instant` constructor is `Instant::now()`. It'd also be nice to have `SystemTime` instances here so that we can do things like have an absolute timeline of events even across multiple, possibly parallel `rustc` processes.

src/profiling_data.rs
wesleywiser: I believe that was the reason I made the change you commented on [here](https://github.com/rust-lang/measureme/pull/5#discussion_r268630912)

wesleywiser:
I like the `Iterator` approach because we get a lot of nice functionality "for free" via the trait and it seems a bit more idiomatic to me.

measureme/src/profiling_data.rs
wesleywiser: Hmm... the link doesn't seem to work right for me so I'm not sure which comment you're referring to.

michaelwoerister: Here's a copy:

We can't do this because the data at `raw_event_bytes` might not have the right alignment. In order to make this safe we have to allocate an event on the stack (e.g. `let mut event = RawEvent::default();`) and then copy the bytes from `raw_event_bytes` into the stack allocated `RawEvent`, i.e. the reverse of this:
https://github.com/rust-lang/measureme/blob/54c175dae6048433106d6099e446c6fd4bb898cd/src/profiler.rs#L59-L73

The important thing is that the `copy_from_slice` calls is done with `u8` slices (which are always safe to work with because of their alignment of 1) instead of pointers to `RawEvent`.

PR #8, opened 2019-03-26
Add end-to-end test for event serialization.

michaelwoerister:
Adds and end-to-end test for event serialization (an adaption of the end-to-end example). This should make it more convenient to implement the `MmapSerializationSink`.


measureme/tests/serialization.rs
```diff
@@ -0,0 +1,108 @@
...
+        count += 1;
+    }
+
+    assert_eq!(count, expected_events.len());
+}
+
+#[test]
+fn test_file_serialization_sink() {
+    let expected_events =
+        generate_profiling_data::<FileSerializationSink>("file_serialization_sink_test");
```
wesleywiser: Do the files generated by this show up when you run `git status` or are they already ignored? If they show up, it'd be great if we could ignore them.

michaelwoerister: Good thinking! I'll take care of that.

PR #16, opened 2019-04-05
Implement a second mmap-based serialization sink that is backed directly by a file.

michaelwoerister:
This might have better performance than the other `mmap`-based sink because it can already write data to the disk while the profiler is still running.

If it does have better performance, I'd like remove the other mmap-sink, I think.

r? @wesleywiser

### DIFF
--- a/measureme/src/async_mmap_serialization_sink.rs
+++ b/measureme/src/async_mmap_serialization_sink.rs
@@ -1,0 +1,108 @@
+use crate::serialization::{Addr, SerializationSink};
+use std::fs::{File, OpenOptions};
+use std::path::{Path};
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::os::unix::io::AsRawFd;
+use std::io;
+
+/// Implements a `SerializationSink` that uses a file-backed mmap.
+pub struct AsyncMmapSerializationSink {
+    file: File,
+    current_pos: AtomicUsize,
+    mapping_start: *mut u8,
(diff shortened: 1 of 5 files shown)


measureme/src/async_mmap_serialization_sink.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl Drop for AsyncMmapSerializationSink {
+    fn drop(&mut self) {
+        let actual_size = *self.current_pos.get_mut();
+
+        unsafe {
+            // First use `mremap` to shrink the memory map. Otherwise `munmap`
+            // would write everything to the backing file, including the
+            // memory we never touched.
+            let new_addr = libc::mremap(self.mapping_start as *mut _,
```
wesleywiser: `mremap` isn't supported on macOS :(

michaelwoerister:
> mremap isn't supported on macOS :(

Hm, that's unfortunate. We need to truncate the mmapped region, otherwise the gigabyte will be written to disk during `munmap`. I guess we'll need to find another way of making this faster.

wesleywiser:
Is it worth using this approach just on Linux? If it's faster than the regular `MmapSerializationSink`, that may help us getting the profiler turned on by default for perf.rlo.

michaelwoerister:
It might be. The benchmark in `benches` shows it to have the same speed as the other `mmap` sink on my system. I never did a real world test case though.

PR #18, opened 2019-04-09
Implement better error handling

### LINKED ISSUE #9: SerializationSink::from_path() should return a Result<>
... so we can handle errors gracefully instead of panicking.

wesleywiser:
Fixes #9

r? @michaelwoerister

### DIFF
--- a/measureme/src/file_serialization_sink.rs
+++ b/measureme/src/file_serialization_sink.rs
@@ -1,4 +1,5 @@
 use crate::serialization::{Addr, SerializationSink};
+use std::error::Error;
 use std::fs;
 use std::io::{BufWriter, Write};
 use std::path::Path;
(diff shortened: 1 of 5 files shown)


measureme/src/file_serialization_sink.rs
```diff
@@ -9,16 +10,14 @@ pub struct FileSerializationSink {
 }
 
 impl SerializationSink for FileSerializationSink {
-    fn from_path(path: &Path) -> Self {
-        // TODO: This is very crude error handling.
-        //       https://github.com/rust-lang/measureme/issues/9
-        fs::create_dir_all(path.parent().unwrap()).unwrap();
+    fn from_path(path: &Path) -> Result<Self, Box<Error>> {
```
wesleywiser: Is `Box<Error>` ok or should we use a better error type?

PR #25, opened 2019-04-12
MmapSerializationSink: Don't panic in Drop::drop().

### LINKED ISSUE #24: Failing to write profiling data to disk should not panic.
The following try-build crashed because it could not write profiling data to disk (insufficient permissions):

https://travis-ci.com/rust-lang/rust/jobs/192490610

Preferably the compiler should not crash/panic because of that. Better to just print a warning to `stderr`. The panic happened in the `Drop` impl of the `MmapSerializationSink`.

michaelwoerister:
Fixes #24.

r? @wesleywiser


michaelwoerister:
Hm, I should also have bumped the version so we can publish to `crates.io`...

wesleywiser:
I can sign off on another PR 🙂

PR #26, opened 2019-04-12
Bump version to 2.0.1

michaelwoerister:
r? @wesleywiser


michaelwoerister:
Hm, why does the commit from #25 still show up?

wesleywiser:
I clicked "Rebase and merge" which apparently *always* rebases even if it's not necessary. The commit hashes are different so git thinks your branch is different even though it has the same content.

PR #28, opened 2019-04-24
Add a stack-collapse tool for generating flamegraphs

wesleywiser:
r? @michaelwoerister


wesleywiser:
That's a great idea! I pushed a commit which adds a basic unit test.

PR #41, opened 2019-05-09
Add a simple file header to binary files created by measureme.

### LINKED ISSUE #40: Add versioning to the binary profile format
Right now the files generated by `rustc -Zself-profile` don't contain any indication of what their concrete encoding is. That should change so that post-processing tools support more than one version or at least give a sensible error message when encountering an unsupported encoding.

For example, each file could start with a some file magic (`b"MMES"` for the event stream, `b"MMSD"` for the string table data, and `b"MMSI"` for the string table index) and a 4 byte little-endian version number that can be used to select the right decoder.

michaelwoerister:
This PR implements a minimal version of file format compatibility checking.

r? @wesleywiser 

Closes #40


measureme/src/file_header.rs
```diff
@@ -0,0 +1,95 @@
...
+        bytes[0 .. 4].copy_from_slice(file_magic);
+        LittleEndian::write_u32(&mut bytes[4..8], CURRENT_FILE_FORMAT_VERSION);
+    });
+}
+
+pub fn read_file_header(
+    bytes: &[u8],
+    expected_magic: &[u8; 4]
+) -> Result<u32, Box<dyn Error>> {
+    debug_assert_eq!(FILE_HEADER_SIZE, 8);
```
Mark-Simulacrum: I'm a bit confused by these asserts -- they don't need to be debug as I expect LLVM to constant-fold them away, and either way I'm not sure what this is intended to do (i.e., because we're comparing against a constant declared in this file). Maybe a comment on the constant that these functions need updating would be better?

michaelwoerister: The assertions are meant to be an extra safeguard against accidentally changing something here. I'd like to leave them in (since it's easy to overlook a comment but not so easy to overlook CI going red). I'll add some comments explaining their purpose though.

michaelwoerister:
I've added some comments and tried to address @Mark-Simulacrum's remarks.

PR #43, opened 2019-05-11
Support wasm and wasi

bjorn3:
This is part of an experiment to port rustc to wasi. (cc https://github.com/rust-lang/miri/issues/722)

Webassembly doesn't support any mmap like mechanism.


bjorn3:
cc @RReverser

PR #50, opened 2019-06-19
[summarize] Implement `diff` support for comparing profiles

### LINKED ISSUE #33: Summarize diffs
When writing a PR to `rustc` you often want either to track the source of a perf regression or to track an improvement. Absolute values don't matter as much.

To that end, it would be nice if you could do `summarize pid-$baseline pid-$change` where `$baseline` could correspond to the `master` branch of `rustc` and `$change` could correspond to your PR which made changes. You could possibly also write `summarize diff ...` instead to simplify the argument parsing logic and infer less.

Example: I wanted this when working on https://github.com/rust-lang/rust/pull/59288; it was hard to work with two summarize outputs side by side.

wesleywiser:
Closes #33

### DIFF
--- a/summarize/src/main.rs
+++ b/summarize/src/main.rs
@@ -3,17 +3,35 @@ extern crate prettytable;
 
 use std::error::Error;
 use std::fs::File;
-use std::io::BufWriter;
+use std::io::{BufReader, BufWriter};
 use std::path::PathBuf;
 use measureme::ProfilingData;
 
 use prettytable::{Table};
+use serde::Serialize;
 use structopt::StructOpt;
 
 mod analysis;
+mod diff;
+mod query_data;
+mod signed_duration;
+
+use query_data::Results;
 
 #[derive(StructOpt, Debug)]
-struct Opt {
+struct DiffOpt {
+    base: PathBuf,
+    change: PathBuf,
+
+    #[structopt(short = "e", long = "exclude")]
+    exclude: Vec<String>,
+
+    #[structopt(long = "json")]
+    json: bool,
+}
+
+#[derive(StructOpt, Debug)]
+struct SummarizeOpt {
     file_prefix: PathBuf,
 
     /// Writes the analysis to a json file next to <file_prefix> instead of stdout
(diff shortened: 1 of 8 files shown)


summarize/src/signed_duration.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if self.is_positive {
+            write!(f, "+")?;
+        } else {
+            write!(f, "-")?;
+        }
+
+        write!(f, "{:?}", self.duration)
+    }
+}
```
bjorn3: missing trailing newline

wesleywiser: fixed

summarize/src/signed_duration.rs
```diff
@@ -0,0 +1,118 @@
...
+        SignedDuration {
+            duration: d,
+            is_positive: true
+        }
+    }
+}
+
+impl Ord for SignedDuration {
+    fn cmp(&self, other: &SignedDuration) -> Ordering {
+        match (self.is_positive, other.is_positive) {
```
Mark-Simulacrum: Maybe easier to change this to `duration.as_nanos() as i128` and then using is_positive to flip the sign if needed?

summarize/src/signed_duration.rs
```diff
@@ -0,0 +1,118 @@
...
+                    (self.duration - rhs.duration, true),
+                (true, true) => (rhs.duration - self.duration, false),
+
+                (false, true) => (self.duration + rhs.duration, false),
+                (true, false) => (self.duration + rhs.duration, true),
+
+                (false, false) if self.duration > rhs.duration =>
+                    (self.duration - rhs.duration, false),
+                (false, false) => (rhs.duration - self.duration, true),
+            };
```
Mark-Simulacrum: I think this would be a lot easier to interpret if we convert to nanoseconds and then back into a Duration.

summarize/src/main.rs
```diff
@@ -25,17 +43,96 @@ struct Opt {
...
+    /// Processes trace files and produces a summary
+    #[structopt(name = "summarize")]
+    Summarize(SummarizeOpt),
+}
+
+fn process_results(file: &PathBuf) -> Result<Results, Box<dyn Error>> {
+    if file.ends_with("json") {
+        let reader = File::open(&file)?;
+
+        let results: Results = serde_json::from_reader(reader)?;
```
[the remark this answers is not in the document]
bjorn3: I did rather go for wrapping it in `BufReader`, so that a file of say 4 gb doesnt get loaded into memory at once. The parsed result is probably smaller, as json contains a lot of special chars, which are not part of the actual parsed content. The [`with_capacity`](https://doc.rust-lang.org/stable/std/io/struct.BufReader.html#method.with_capacity) method allows building with a large capacity, like 128mb. Using a `BufReader` can also reduce latency, as parsing can already start while the file isnt fully read yet.

summarize/src/signed_duration.rs
wesleywiser: That's a really good idea! That simplifies this a lot.

summarize/src/main.rs
```diff
@@ -25,17 +43,96 @@ struct Opt {
...
+    } else {
+        let data = ProfilingData::new(&file)?;
+
+        Ok(analysis::perform_analysis(data))
+    }
+}
+
+fn write_results_json(file: &PathBuf, results: impl Serialize) -> Result<(), Box<dyn Error>> {
+    let file = BufWriter::new(File::create(file.with_extension("json"))?);
+    serde_json::to_writer(file, &results)?;
```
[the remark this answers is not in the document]
wesleywiser: I think for now, I'd prefer to leave the standard `BufReader`\`BufWriter` alone. These tools run pretty quick already so I don't think this is going to make a big difference.

wesleywiser:
I think this is ready to merge. @Mark-Simulacrum is that ok with you?

Mark-Simulacrum:
Feel free to merge when ready -- I forget if we have bors or anything setup here so won't do it myself in this case

PR #57, opened 2019-09-24
Store some metadata about the profile in a reserved string table slot

### LINKED ISSUE #15: Store some metadata about the profile in a reserved string table slot.
It might be useful to store data about a collected profile with the profile itself (like process-id, start-time). It occurred to me that we could just pack this data into a JSON string and store that under a pre-reserved `StringId` in the string table. Then we don't need to implement this in the more cumbersome binary format. 

cc @wesleywiser

wesleywiser:
Fixes #15

### DIFF
--- a/measureme/src/profiler.rs
+++ b/measureme/src/profiler.rs
@@ -1,7 +1,7 @@
 use crate::file_header::{write_file_header, FILE_MAGIC_EVENT_STREAM};
 use crate::raw_event::{RawEvent, Timestamp, TimestampKind};
 use crate::serialization::SerializationSink;
-use crate::stringtable::{SerializableString, StringId, StringTableBuilder};
+use crate::stringtable::{METADATA_STRING_ID, SerializableString, StringId, StringTableBuilder};
 use std::error::Error;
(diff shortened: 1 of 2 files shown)


wesleywiser:
Fixed and added some additional docs.

measureme/src/profiler.rs
```diff
@@ -42,11 +42,26 @@ impl<S: SerializationSink> Profiler<S> {
...
+        };
+
+        let mut args = String::new();
+        for arg in std::env::args() {
+            args.push_str(&arg.escape_default().to_string());
+            args.push(' ');
+        }
+
+        profiler.string_table.alloc_metadata(&*format!(
+            "{{ \"start_time\": {}, \"process_id\": {}, \"cmd\": \"{}\" }}",
```
michaelwoerister: This might be a perfect place for using raw string literals.
https://rahul-thakoor.github.io/rust-raw-string-literals/

wesleywiser: That's a great idea.

amanjeev:
Ah! I had forgotten. :( Thank you @wesleywiser

PR #58, opened 2019-09-25
correct fault in collapse-threads command

andjo403:
found that I hade made an assumption that the thread_id stated the order of the start times per thread but that is not the case threads can be started long before the first event


wesleywiser:
Sorry for the late review! The changes look fine to me but I'm struggling to come up with a case that this makes a difference for. Can you give an example of when this will produce a different result?

andjo403:
I noticed that I got more "threads" then expected when I did a release build for crox due to the codegen_run_lto started an thread before all the optimize threads. this resulted in that the codegen_run_lto thread take one of the first end times that is better to use for one of the optimize threads. so the problem is that not all the threads is collapsed.

PR #59, opened 2019-09-30
use complete events in crox

### LINKED ISSUE #47: Reduce Chrome profiler file size by using Complete events
Right now, the `crox` tool emits one `Begin` event and one `End` event for each non-instant event. The `Complete` event combines the `Begin` and `End` events together which should substantially reduce the size of the json file (up to half). 

[Chrome Profiler File Format Docs](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit#heading=h.lpfof2aylapb)

andjo403:
to reduce the size of the generated json file
fixes #47

### DIFF
--- a/crox/src/main.rs
+++ b/crox/src/main.rs
@@ -5,7 +5,7 @@ use std::path::PathBuf;
 use std::time::Duration;
 use std::time::SystemTime;
 
-use measureme::{ProfilingData, TimestampKind};
+use measureme::{MatchingEvent, ProfilingData};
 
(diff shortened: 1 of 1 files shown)


andjo403:
@wesleywiser you hade made it very easy to do with the iterator.
the crate that I looked at when I wanted this was script from servo where the file went from 1.4 GB to 0.7 GB and that was enough to get chrome to show the file after 5 min and using 6 GB memory :)

PR #60, opened 2019-09-30
add option to crox to filter out short events

andjo403:
some files is to large to show in the browser
filter_duration_ms filter out all events that is shorter than specified limit


andjo403:
changed to microseconds was I that not thought last night

PR #70, opened 2019-10-24
Add RAII-based API for recording events and bump version to 0.4.0

michaelwoerister:
This PR adds a RAII-based API for recording events in preparation for replacing `(start, end)` event pairs with single "interval" events. It also bumps the version to 0.4.0 so the new API can be used in `rustc`. 

r? @wesleywiser

### DIFF
--- a/measureme/src/profiler.rs
+++ b/measureme/src/profiler.rs
@@ -78,6 +78,8 @@ impl<S: SerializationSink> Profiler<S> {
         self.string_table.alloc(s)
     }
 
+    /// Records an event with the given parameters. The event time is computed
+    /// automatically.
     pub fn record_event(
         &self,
         event_kind: StringId,
@@ -90,16 +92,16 @@ impl<S: SerializationSink> Profiler<S> {
             + duration_since_start.subsec_nanos() as u64;
         let timestamp = Timestamp::new(nanos_since_start, timestamp_kind);
 
+        let raw_event = RawEvent {
+            event_kind,
+            id: event_id,
+            thread_id,
+            timestamp,
+        };
+
         self.event_sink
             .write_atomic(std::mem::size_of::<RawEvent>(), |bytes| {
                 debug_assert_eq!(bytes.len(), std::mem::size_of::<RawEvent>());
-
-                let raw_event = RawEvent {
-                    event_kind,
-                    id: event_id,
-                    thread_id,
-                    timestamp,
-                };
 
                 let raw_event_bytes: &[u8] = unsafe {
                     std::slice::from_raw_parts(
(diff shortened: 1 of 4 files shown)


measureme/src/profiler.rs
```diff
@@ -111,4 +113,46 @@ impl<S: SerializationSink> Profiler<S> {
                 bytes.copy_from_slice(raw_event_bytes);
             });
     }
+
+    /// Creates a "start" event and returns a `TimingGuard` that will create
+    /// the corresponding "end" event when it is dropped.
+    pub fn start_event<'a>(
```
michaelwoerister: Is there a better name for this? `start_recording` maybe?

measureme/src/profiler.rs
```diff
@@ -111,4 +113,46 @@ impl<S: SerializationSink> Profiler<S> {
...
+            event_id,
+            event_kind,
+            thread_id,
+        })
+    }
+}
+
+/// When dropped, this `TimingGuard` will record an "end" event in the
+/// `Profiler` it was created by.
+#[must_use]
```
bjorn3: Why must_use? That would mean that you would have to manually call `std::mem::drop` to prevent a compiler warning.

michaelwoerister: `let _guard = ...` also works without a warning. This code is basically copied from what we already have in `rustc`:
https://github.com/rust-lang/rust/blob/8e0007f829661e57d008d2e908c95f6e84b04b25/src/librustc/util/profiling.rs#L348-L349

It works well there.

michaelwoerister: And in the compiler version the `#[must_use]` attribute has prevented me from introducing at least two bugs, so I think it's useful `:)`

bjorn3: Didnt know that.

measureme/src/profiler.rs
michaelwoerister: Is `start_recording_interval_event` too wordy?

wesleywiser: `start_recording` implies to me that if you don't call it, any calls to `record_event` will not be recorded.

I think this is fine for now.

measureme/src/profiler.rs
```diff
@@ -111,4 +113,46 @@ impl<S: SerializationSink> Profiler<S> {
...
+pub struct TimingGuard<'a, S: SerializationSink>(TimingGuardInternal<'a, S>);
+
+struct TimingGuardInternal<'a, S: SerializationSink> {
```
michaelwoerister: I just realized that the `TimingGuard`/`TimingGuardInternal` split is not necessary for measureme. Let me fix that...

measureme/src/profiler.rs
wesleywiser: We could eventually split the api more and hide `record_event` and just expose two different methods `record_duration` and `record_instantaneous`.

wesleywiser:
Are there changes we should make to `RawEvent` before we ship 0.4?

measureme/src/profiler.rs
michaelwoerister: I changed it to `start_recording_interval_event` just now, and plan to later replace `record_event` with `record_instant_event`. Does that work for you?

michaelwoerister:
> Are there changes we should make to `RawEvent` before we ship 0.4?

We could change ThreadId to `u32` but the rest of the changes depend on actually switching to "interval" events, I'd say.

michaelwoerister:
But let's modify `RawEvent` just now. None of the changes above affect `summarize` or other tools, so we can bump `rustc` to 0.4.0 without worrying about perf.rlo just now.

measureme/src/profiler.rs
wesleywiser: Yeah, that's fine.

wesleywiser:
> But let's not modify RawEvent just yet. None of the changes above affect summarize or other tools, so we can bump rustc to 0.4.0 without worrying about perf.rlo.

I agree. When we get to that point, we'll have to make sure we loop in @Mark-Simulacrum so we don't break perf.rlo.

PR #72, opened 2019-11-01
measureme: Add a ProfilingDataBuilder that can be used for constructing test data.

michaelwoerister:
This is useful because data generated from an actual `Profiler` contains unpredictable timestamps.

r? @wesleywiser

### DIFF
--- a/measureme/src/profiling_data.rs
+++ b/measureme/src/profiling_data.rs
@@ -1,10 +1,18 @@
 use crate::event::Event;
-use crate::file_header::FILE_HEADER_SIZE;
-use crate::{ProfilerFiles, RawEvent, StringTable, TimestampKind};
+use crate::file_header::{
+    read_file_header, write_file_header, CURRENT_FILE_FORMAT_VERSION, FILE_HEADER_SIZE,
+    FILE_MAGIC_EVENT_STREAM,
+};
+use crate::serialization::ByteVecSink;
+use crate::{
+    ProfilerFiles, RawEvent, SerializationSink, StringTable, StringTableBuilder, Timestamp,
+    TimestampKind,
+};
 use std::error::Error;
 use std::fs;
 use std::mem;
 use std::path::Path;
+use std::sync::Arc;
 use std::time::{Duration, SystemTime};
 
 pub struct ProfilingData {
@@ -20,6 +28,15 @@ impl ProfilingData {
         let index_data =
             fs::read(paths.string_index_file).expect("couldn't read string_index file");
         let event_data = fs::read(paths.events_file).expect("couldn't read events file");
+
+        let event_data_format = read_file_header(&event_data, FILE_MAGIC_EVENT_STREAM)?;
+        if event_data_format != CURRENT_FILE_FORMAT_VERSION {
+            Err(format!(
+                "Event stream file format version '{}' is not supported
+                 by this version of `measureme`.",
+                event_data_format
+            ))?;
+        }
 
         let string_table = StringTable::new(string_data, index_data)?;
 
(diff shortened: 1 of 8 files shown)


michaelwoerister:
I haven't found the time to write actual test cases with this yet (e.g. for `summarize`). Maybe later today.

michaelwoerister:
Beware! `dbg!` strikes again. Let me fix that.

michaelwoerister:
> Beware! `dbg!` strikes again. Let me fix that.

Fixed.

michaelwoerister:
I added a single test case to `summarize` just now. That's all I'm going to do for this PR, so feel free to review any time.

measureme/src/profiling_data.rs
```diff
@@ -21,6 +29,15 @@ impl ProfilingData {
             fs::read(paths.string_index_file).expect("couldn't read string_index file");
         let event_data = fs::read(paths.events_file).expect("couldn't read events file");
 
+        let event_data_format = read_file_header(&event_data, FILE_MAGIC_EVENT_STREAM)?;
+        if event_data_format != CURRENT_FILE_FORMAT_VERSION {
```
wesleywiser: Do we need to support `CURRENT_FILE_FORMAT_VERSION - 1` for perf.rlo?

cc @Mark-Simulacrum

measureme/src/profiling_data.rs
```diff
@@ -152,3 +169,234 @@ impl<'a> Iterator for MatchingEventsIterator<'a> {
...
+
+/// A `ProfilingDataBuilder` allows for programmatically building
+/// `ProfilingData` objects. This is useful for writing tests that expect
+/// `ProfilingData` with predictable events (and especially timestamps) in it.
+///
+/// `ProfilingDataBuilder` provides a convenient interface but its
+/// implementation might not be efficient, which why it should only be used for
+/// writing tests and other things that are not performance sensitive.
+pub struct ProfilingDataBuilder {
+    event_sink: Arc<ByteVecSink>,
```
wesleywiser: Do we need to use `Arc`s here? Unit tests should be single threaded so we don't need this to work across multiple threads.

measureme/src/profiling_data.rs
Mark-Simulacrum: I imagine that we don't care too much in practice, we just want me to be aware when we make changes -- as it currently stands, we don't really support this too well, but we can either just deal with commits failing or some other way to migrate to a new version. Once we're including measureme in the artifacts rustc CI produces that would be helpful as well.

wesleywiser: Ok. We'll be sure to coordinate with you when we bump rustc's dependency version to the one with breaking changes.

measureme/src/profiling_data.rs
michaelwoerister: You're right. I basically copied this from the string table case, but here a plain EventSink should work just fine.

michaelwoerister: I removed the `Arc`.

michaelwoerister:
@wesleywiser I implemented your suggestion and add more test cases. This is ready for review again.

summarize/src/analysis.rs
```diff
@@ -164,3 +164,255 @@ pub fn perform_analysis(data: ProfilingData) -> Results {
...
+        assert_eq!(results.query_data_by_label("q1").invocation_count, 1);
+        assert_eq!(results.query_data_by_label("q2").invocation_count, 1);
+        assert_eq!(results.query_data_by_label("q3").invocation_count, 1);
+    }
+
+    #[test]
+    fn events_with_same_starting_time() {
+        //                      <--e4-->
+        //                      <---e3--->
+        //  <--------e1--------><--------e2-------->
```
wesleywiser: These are super helpful for understanding the tests. Thanks for adding them!

michaelwoerister: Yes, they are super helpful for writing the tests too `:)`

Issue #42, opened 2019-05-10
Integrated FlameGraph output


wesleywiser:
We already have a tool to generate collapsed stacks from profile data. However, this still needs to be piped into the FlameGraph script to generate svg output. We should integrate [inferno](https://github.com/jonhoo/inferno) to generate the svg without needing external scripts.

michaelwoerister:
Nice! Yes, please.

PR #73, opened 2019-11-05
use Inferno to generate flamegraphs

andjo403:
closes https://github.com/rust-lang/measureme/issues/42


bjorn3:
Keeping the ability to get the raw stack collapsed data is useful. You for example need them to create differential flamegraphs. It is also useful as input for other programs.

andjo403:
shall I add a new crate toolsLib or something like that? and move the measureme/flamegraph/src/stack_collapse.rs there (later maybe also measureme/src/profiling_data.rs can be moved there) and then create the new flamegraph crate.

michaelwoerister:
This looks awesome, thanks @andjo403!

wesleywiser:
@andjo403 Yeah, I think I gave you bad advice on Zulip. Sorry about that! 

I agree with @bjorn3 that `stack_collapse` is useful by itself so we shouldn't remove it.

> create the new flamegraph crate

I think this is the best thing to do. Adding a new crate to the workspace will also let you reuse code from `stack_collapse` inside it since we shouldn't duplicate that logic.

andjo403:
crated the new crates tools_lib and flamegraph

PR #74, opened 2019-11-05
stack collaps events from multiple threads correct

andjo403:
part of the problem seen in https://github.com/rust-lang/rust/issues/65788
most of the remaining unaccounted for time I think is the main threads wait for the link to start after codegen_crate is done.
have compared the times from crox with the per thread data in the flamegraph and they match with this PR

### DIFF
--- a/flamegraph/src/main.rs
+++ b/flamegraph/src/main.rs
@@ -2,7 +2,6 @@ use std::error::Error;
 use std::fs::File;
 use std::io::BufWriter;
 use std::path::PathBuf;
-use std::time::Duration;
 
 use measureme::ProfilingData;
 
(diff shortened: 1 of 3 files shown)


stack_collapse/src/stack_collapse.rs
```diff
@@ -5,38 +5,33 @@ use measureme::{Event, TimestampKind};
 
 pub fn collapse_stacks<'a>(
     events: impl Iterator<Item = Event<'a>>,
-    first_event_time: SystemTime,
```
wesleywiser: Ah, so the bug is basically that we should have been tracking the "first event time" of each thread individually instead of looking at the first event across all threads? The effect of that would make sense given the reports I've seen; the longer the crate takes to get to codegen (at which point, multiple threads come into play), the more unaccounted for time that was being reported.

andjo403: yes but also track the next_observation_time need to be tracked per thread as there can be out of order events between threads

andjo403:
have rebased now

PR #76, opened 2019-11-06
Replace (start, stop) event pairs with interval events.

michaelwoerister:
This reduces the number of events that need to be recorded.

~~Currently the PR updates all tools except `stack_collapse` which seems to be in flux at the moment. It also does not yet use the raw event format (i.e. the number of bits used for timestamps and thread-ids is unnecessarily high).~~

I intend to optimize the raw format of events and the string table in subsequent PR. This one already contains enough changes.

@wesleywiser Let me know if prefer the PR to be split up into more commits.

### DIFF
--- a/crox/src/main.rs
+++ b/crox/src/main.rs
@@ -5,9 +5,10 @@ use std::path::PathBuf;
 use std::time::Duration;
 use std::time::SystemTime;
 
-use measureme::{MatchingEvent, ProfilingData};
+use measureme::{ProfilingData, Timestamp};
 
 use serde::{Serialize, Serializer};
+use std::cmp;
 use structopt::StructOpt;
 
 fn as_micros<S: Serializer>(d: &Duration, s: S) -> Result<S::Ok, S::Error> {
@@ -64,14 +65,12 @@ fn generate_thread_to_collapsed_thread_mapping(
         for event in data.iter() {
             thread_start_and_end
                 .entry(event.thread_id)
-                .and_modify(|(start, end)| {
-                    if *start > event.timestamp {
-                        *start = event.timestamp;
-                    } else if *end < event.timestamp {
-                        *end = event.timestamp;
-                    }
+                .and_modify(|(thread_start, thread_end)| {
+                    let (event_min, event_max) = timestamp_to_min_max(event.timestamp);
+                    *thread_start = cmp::min(*thread_start, event_min);
+                    *thread_end = cmp::max(*thread_end, event_max);
                 })
-                .or_insert_with(|| (event.timestamp, event.timestamp));
+                .or_insert_with(|| timestamp_to_min_max(event.timestamp));
         }
         // collect the the threads in order of the end time
         let mut end_and_thread = thread_start_and_end
@@ -120,37 +119,34 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     //find the earlier timestamp (it should be the first event)
     //subtract one tick so that the start of the event shows in Chrome
-    let first_event_timestamp = data.iter().next().unwrap().timestamp - Duration::from_micros(1);
+    let first_event_timestamp = make_start_timestamp(&data);
 
     let mut serializer = serde_json::Serializer::new(chrome_file);
     let thread_to_collapsed_thread = generate_thread_to_collapsed_thread_mapping(&opt, &data);
-    let mut event_iterator = data.iter_matching_events();
+    let mut event_iterator = data.iter();
 
     //create an iterator so we can avoid allocating a Vec with every Event for serialization
     let json_event_iterator = std::iter::from_fn(|| {
         while let Some(event) = event_iterator.next() {
             // Chrome does not seem to like how many QueryCacheHit events we generate
             // only handle startStop events for now
-            if let MatchingEvent::StartStop(start, stop) = event {
-                let duration = stop.timestamp.duration_since(start.timestamp).unwrap();
+            if let Timestamp::Interval { start, end } = event.timestamp {
+                let duration = end.duration_since(start).unwrap();
                 if let Some(minimum_duration) = opt.minimum_duration {
                     if duration.as_micros() < minimum_duration {
                         continue;
                     }
                 }
                 return Some(Event {
-                    name: start.label.clone().into_owned(),
-                    category: start.event_kind.clone().into_owned(),
+                    name: event.label.clone().into_owned(),
+                    category: event.event_kind.clone().into_owned(),
                     event_type: EventType::Complete,
-                    timestamp: start
-                        .timestamp
-                        .duration_since(first_event_timestamp)
-                        .unwrap(),
+                    timestamp: start.duration_since(first_event_timestamp).unwrap(),
                     duration,
                     process_id: 0,
                     thread_id: *thread_to_collapsed_thread
-                        .get(&start.thread_id)
-                        .unwrap_or(&start.thread_id),
+                        .get(&event.thread_id)
+                        .unwrap_or(&event.thread_id),
                     args: None,
                 });
             }
@@ -162,4 +158,34 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     serializer.collect_seq(json_event_iterator)?;
 
     Ok(())
+}
+
+fn timestamp_to_min_max(timestamp: Timestamp) -> (SystemTime, SystemTime) {
+    match timestamp {
+        Timestamp::Instant(t) => (t, t),
+        Timestamp::Interval { start, end } => {
+            // Usually start should always be greater than end, but let's not
+            // choke on invalid data here.
+            (cmp::min(start, end), cmp::max(start, end))
+        }
+    }
+}
+
+// FIXME: Move this to `ProfilingData` and base it on the `start_time` field
+//        from metadata.
+fn make_start_timestamp(data: &ProfilingData) -> SystemTime {
+    // We cannot assume the first event in the stream actually is the first
+    // event because interval events are emitted at their end. So in theory it
+    // is possible that the event with the earliest starting time is the last
+    // event in the stream (i.e. if there is an interval event that spans the
+    // entire execution of the profile).
+    //
+    // Let's be on the safe side and iterate the whole stream.
+    let min = data
+        .iter()
+        .map(|e| timestamp_to_min_max(e.timestamp).0)
+        .min()
+        .unwrap();
+
+    min - Duration::from_micros(1)
 }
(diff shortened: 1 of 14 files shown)


crox/src/main.rs
```diff
@@ -163,3 +159,31 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    //
+    // Let's be on the save side and iterate the whole stream.
+    let min = data
+        .iter()
+        .map(|e| timestamp_to_min_max(e.timestamp).0)
+        .min()
+        .unwrap();
+
+    min - Duration::from_micros(1)
+}
```
andjo403: I think this shall be moved to profiling_data.rs and in the future use the start_time from the metadata header

crox/src/main.rs
```diff
@@ -163,3 +159,31 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     Ok(())
 }
+
+fn timestamp_to_min_max(timestamp: Timestamp) -> (SystemTime, SystemTime) {
+    match timestamp {
+        Timestamp::Instant(t) => (t, t),
```
andjo403: hmm maybe Instants shall not be handled and filtered out before calling this as it is not part of the output

summarize/src/analysis.rs
```diff
@@ -20,141 +74,76 @@ pub fn perform_analysis(data: ProfilingData) -> Results {
...
+                            "Ignoring event with unknown event kind `{}`",
+                            unknown_event_kind
+                        );
+                    }
+                };
+
+                // Update the total_time counter
+                let current_event_is_top_level = thread_stack.is_empty();
+                if current_event_is_top_level {
+                    total_time += current_event.duration().unwrap();
```
andjo403: think that this is missing alot of time hade prefered if the first and last time for the thread/or the whole log was used

andjo403: in crox files I can see multiple seconds before and after analysis that is not covered by any event maybe there shall be an event started in the main function that cover everything

crox/src/main.rs
michaelwoerister: Seems like a good idea.

summarize/src/analysis.rs
michaelwoerister: Yeah, it's probably a good idea to measure both, the sum of top-level event times, and the sum of min/max times of all threads. Plus the wall-time.

In the last working group meeting we mentioned that we want to add a generic key-value map to the summarize output. The different "total time" measurements (and an estimate of time not accounted for) could go in there.

crox/src/main.rs
michaelwoerister: I'll leave a comment. I don't want to do it as part of this PR as it would require adding metadata decoding too and I want to keep this a refactoring PR as much as possible.

crox/src/main.rs
michaelwoerister: Maybe. I'll leave this as is right now, as it matches the current behavior.

measureme/src/profiling_data.rs
```diff
@@ -94,79 +90,54 @@ impl<'a> Iterator for ProfilerEventIterator<'a> {
 
         let string_table = &self.data.string_table;
 
-        let mut timestamp = SystemTime::UNIX_EPOCH;
-        timestamp += Duration::from_nanos(raw_event.timestamp.nanos());
+        // FIXME: Lots of Rust code compiled in the seventies apparently.
```
wesleywiser: 😆

measureme/src/profiling_data.rs
```diff
@@ -306,26 +265,50 @@ impl ProfilingDataBuilder {
     }
 }
 
+impl<'a> ExactSizeIterator for ProfilerEventIterator<'a> {}
+
+fn event_index_to_addr(event_index: usize) -> usize {
```
wesleywiser: Is this being used?

crox/src/main.rs
```diff
@@ -163,3 +159,33 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+// FIXME: Move this to `ProfilingData` and base it on the `start_time` field
+//        from metadata.
+fn make_start_timestamp(data: &ProfilingData) -> SystemTime {
+    // We cannot assume the first event in the stream actually is the first
+    // event because interval events are emitted at their end. So in theory it
+    // is possible that the event with the earliest starting time is the last
+    // event in the stream (i.e. if there is an interval event that spans the
+    // entire execution of the profile).
+    //
+    // Let's be on the save side and iterate the whole stream.
```
wesleywiser: nit: `safe`

summarize/src/analysis.rs
```diff
@@ -20,141 +74,76 @@ pub fn perform_analysis(data: ProfilingData) -> Results {
...
-                        data.self_time += duration;
-                        data.number_of_cache_misses += 1;
-                        data.invocation_count += 1;
+            Timestamp::Interval { .. } => {
+                // This is an interval event
+                let thread_stack = threads.entry(current_event.thread_id).or_default();
+
+                // Pop all events from the stack that are not parents of the
+                // current event.
+                while let Some(current_top) = thread_stack.last().cloned() {
```
wesleywiser: Why would we have events on the stack that aren't parents of the current event?

michaelwoerister: Since we don't explicit "end" events anymore, this is now the only mechanism for popping things. Consider the following situation:
```
     <--e2-->   <--e3-->
 <-----------e1----------->
```
In the event stream this shows up as something like:
```
[
	{ label=e2, start= 5, end=10 },
	{ label=e3, start=15, end=20 },
	{ label=e1, start= 0, end=25 },
]
```
because events are emitted in the order of the end timestamps.  So, as we walk backwards, we

1. encounter `e1`, push it onto the stack, then
2. encounter `e3`, the stack contains `e1`, but that is fine since the time-interval of `e1` includes the time interval of `e3`. `e3` goes onto the stack and then we
3. encounter `e2`. The stack is `[e1, e3]`, but now `e3` needs to be popped because we are past its range, so we pop `e3` and push `e2`.

Why is popping done in a `while` loop? consider the following

```
                 <-e4->
     <--e2-->   <--e3-->
 <-----------e1----------->
```

This looks as follows in the stream:
```
[
	{ label=e2, start= 5, end=10 },
	{ label=e4, start=17, end=19 },
	{ label=e3, start=15, end=20 },
	{ label=e1, start= 0, end=25 },
]
```
In this case when we encounter `e2`, the stack is `[e1, e3, e4]`, and both `e4` and `e3` need to be popped in the same step.

measureme/src/profiling_data.rs
michaelwoerister: Yes, in `decode_event`. It was used multiple times in a previous version but not anymore it seems. I'll inline it.

wesleywiser: Nah, it's fine. I just couldn't find where it was used.

summarize/src/analysis.rs
wesleywiser: Brilliant, thanks!

michaelwoerister: I'll see if I can add the above comment to the docs somewhere.

tools_lib/src/stack_collapse.rs
```diff
@@ -1,183 +1,215 @@
...
+        }
 
-            while event.timestamp > *next_observation_time {
-                *count += 1;
-                *next_observation_time += Duration::from_millis(interval);
+        let sampling_interval_nanos = sampling_interval.as_nanos();
+
+        let thread_duration_nanos = max_sampled_timestamp.duration_since(thread_start_time).unwrap().as_nanos();
+        let samples = thread_duration_nanos / sampling_interval_nanos;
+        thread_start_time + Duration::from_nanos((samples * sampling_interval_nanos).try_into().unwrap())
```
michaelwoerister: Doing all this logic here might be overkill. We could also start sampling at 
- `last_event.end()`, which would mean there is always one sampling point with stack `rustc`, or
- `last_event.end() - Duration::from_nanos(1)`, which would mean that the earliest sampling point is not exactly at the beginning of the earliest event.

Either case should be fine for real world profiling data. Thoughts?

tools_lib/src/stack_collapse.rs
```diff
@@ -1,183 +1,215 @@
...
+pub fn collapse_stacks<'a>(
+    profiling_data: &ProfilingData,
+    sampling_interval: Duration,
+) -> HashMap<String, u64> {
+    let thread_ids = find_all_thread_ids_and_start_times(profiling_data);
+    let mut counters = HashMap::new();
+
+    for (thread_id, thread_start_time) in thread_ids {
+        process_thread(thread_id, thread_start_time, profiling_data, sampling_interval, &mut counters);
+    }
```
michaelwoerister: This implementation iterators the profiling data once for each thread and once in the beginning for finding all thread ids. It think this makes the code quite a bit simpler. An alternative approach that is exactly as simple but trades processing time for memory consumption is to iterate once and building up a vec of events for each thread.

tools_lib/src/stack_collapse.rs
```diff
@@ -1,183 +1,215 @@
...
+        });
+
+        let profiling_data = b.into_profiling_data();
+
+        let recorded_stacks = super::collapse_stacks(&profiling_data, Duration::from_nanos(1));
+
+        let mut expected_stacks = HashMap::<String, u64>::new();
+        expected_stacks.insert("rustc;EventB;EventA".into(), 200);
+        expected_stacks.insert("rustc;EventB".into(), 200);
+        expected_stacks.insert("rustc;EventA".into(), 200);
```
michaelwoerister: Still need to rebase this onto https://github.com/rust-lang/measureme/pull/78...

andjo403:
pushed an alternative collapse algorithm to https://github.com/michaelwoerister/measureme/pull/1
you can take parts of it if you like e.g. the tests

tools_lib/src/stack_collapse.rs
andjo403: think that is is overkill to track the thread start so in this case I prefere `last_event.end() - Duration::from_nanos(1)`

tools_lib/src/stack_collapse.rs
andjo403: a bit worried about the time for e.g when compiling servo script and using incremental there will be around 1500 threads and to iterate that many times over all events feels like it will take forever

tools_lib/src/stack_collapse.rs
andjo403: you can take it from michaelwoerister#1

michaelwoerister: Took the whole algorithm from michaelwoerister#1 `:D`

summarize/src/analysis.rs
michaelwoerister: I switched to the method you proposed, @andjo403.

michaelwoerister:
OK, I think this is ready for the final review, @wesleywiser. I addressed all the comments given so far.

PR #78, opened 2019-11-06
add description to stack collapse tests

andjo403:


tools_lib/src/stack_collapse.rs
```diff
@@ -56,116 +56,170 @@ mod test {
 
     #[test]
     fn basic_test() {
+        //                                         <--e1-->
```
michaelwoerister: Should this be `e3`?

tools_lib/src/stack_collapse.rs
```diff
@@ -56,116 +56,170 @@ mod test {
...
+        //                 <--e1-->        <----------e2---------->
+        //              T2 1       2       3       4       5       6
+        // sample interval |   |   |   |   |   |   |   |   |   |   |
+        // stacks count:
+        // rustc                       1   2
+        // rustc;e1            1   2
+        // rustc;e2                            1   2           3   4
+        // rustc;e2;e3                                 1   2
```
andjo403: think this shall be e1 want to use the same event to show that it is on the stack here
```suggestion
        // rustc;e2;e1                                 1   2
```

tools_lib/src/stack_collapse.rs
andjo403: see my comment below

tools_lib/src/stack_collapse.rs
michaelwoerister: :+1:

andjo403:
fixed the fault

Issue #77, opened 2019-11-06
Tools should use FxHashMap instead sip-hash based hash maps or BTreeMaps


michaelwoerister:
`FxHashMap` from `rustc-hash` is fast and we are doing lots of hashing.

silathdiir:
Hello, I'm interested in working on this issue.

@michaelwoerister , I did a search on the source code. Should I replace all calling `std::collections::HashMap` and `std::collections::BTreeMap` with `FxHashMap`?

Please correct me if I was wrong. Thank you.

michaelwoerister:
Yes, I think all HashMaps and BTreeMaps can be replaced with FxHashMap. It would be good to wait until https://github.com/rust-lang/measureme/pull/76 has landed though, as that PR changes a lot of code.

silathdiir:
Got it. Thank you. I will move on until #76 is merged.

michaelwoerister:
#76 has just been merged.

silathdiir:
Hi @michaelwoerister , I tries to fix this on PR #82 . Please help review it.

And I am sorry that I could not find the corresponding method `HashMap::with_capacity` in FxHashMap (and FxHashSet).
I used `FxHashMap::default` directly. I could not confirm if it is right. Please correct me if i was wrong.

silathdiir:
Based on @michaelwoerister 's help, I fixed to replace `HashMap::with_capacity` with `FxHashMap::with_capacity_and_hasher`. Thanks.

michaelwoerister:
Closing. Implemented in #82.

PR #82, opened 2019-11-08
Replace HashMap with `rustc_hash::FxHashMap` in Tools

silathdiir:
### Summary

Tries to fix Issue #77 .

1. Replaces HashMap with `rustc_hash::FxHashMap`.
2. Replaces HashSet with `rustc_hash::FxHashSet`.

Based on @michaelwoerister 's help,  I fixed to replace `HashMap::with_capacity` with `FxHashMap::with_capacity_and_hasher`. Thanks.

### DIFF
--- a/summarize/src/diff.rs
+++ b/summarize/src/diff.rs
@@ -1,7 +1,7 @@
 use crate::query_data::{QueryData, QueryDataDiff, Results};
 use crate::signed_duration::SignedDuration;
+use rustc_hash::{FxHashMap, FxHashSet};
 use serde::{Deserialize, Serialize};
-use std::collections::{HashMap, HashSet};
 use std::time::Duration;
 
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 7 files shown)


summarize/src/diff.rs
```diff
@@ -10,8 +10,8 @@ pub struct DiffResults {
     pub total_time: SignedDuration,
 }
 
-fn build_query_lookup(query_data: &[QueryData]) -> HashMap<&str, usize> {
-    let mut lookup = HashMap::with_capacity(query_data.len());
+fn build_query_lookup(query_data: &[QueryData]) -> FxHashMap<&str, usize> {
+    let mut lookup = FxHashMap::default();
```
wesleywiser: `FxHashMap` is just a type alias for the standard library's `HashMap` so you should be able to call the same function like this `FxHashMap::with_capacity(query_data.len())`.

michaelwoerister: Unfortunately that doesn't work because `with_capacity()` hardcodes the default `Hasher`. Instead we have to use the slight more verbose `FxHashMap::with_capacity_and_hasher(query_data.len(), Default::default())`.

silathdiir: Thanks. Will fix it.

PR #83, opened 2019-11-09
break out read log part of measureme to the tools_lib crate

andjo403:


andjo403:
not super happy about the solution with having the testing_common.rs in the tools_lib as it feels like it is only for tests in measureme. one possibility is to move out the bench and test folder from measureme but it feels like they belong in measureme

wesleywiser:
@michaelwoerister would you like to review as well? This moves the code which reads profiling data into the `tools_lib` crate which is nice because it allows us to add helpful dependencies such as `serde` without increasing the size of the `measureme` crate.

measureme/Cargo.toml
```diff
@@ -18,3 +18,6 @@ rustc-hash = "1.0.1"
 
 [target.'cfg(not(target_arch="wasm32"))'.dependencies]
 memmap = "0.6.0"
+
+[dev-dependencies]
+tools_lib = { path = "../tools_lib" }
```
michaelwoerister: I'd like to get rid of this (pseudo)circular dependency. I think it's fine to move the end-to-end serialization tests into `tools_lib`.

tools_lib/src/lib.rs
```diff
@@ -1 +1,21 @@
...
+//! # Reading event trace files
+//!
+//! The main entry point for reading trace files is the [`ProfilingData`] struct.
+//!
+//! To create a [`ProfilingData`], call the [`ProfilingData::new()`] function and provide a `Path` with the directory and file name for the trace files.
+//!
+//! To retrieve an `Iterator` of all of the events in the file, call the [`ProfilingData::iter()`] method.
+//!
+//! [`ProfilingData`]: struct.ProfilingData.html
+//! [`ProfilingData::iter()`]: struct.ProfilingData.html#method.iter
```
michaelwoerister: This doc string could use some line breaks.

andjo403:
fixed the comments

michaelwoerister:
Thanks, @andjo403!

PR #84, opened 2019-11-09
use meta data in crox events to be able to merge multiple logs

andjo403:
add parameter `--dir` that merges all events in the directory to one chrome_profiler.json file.
makes it possible to give multiple file_prefixes and will merge all of them to one chrome_profiler.json file.
also use the crate name from the cmd metadata to show what process_id  is from what crate build.


wesleywiser:
Some screenshots from doing `RUSTFLAGS="-Z self-profile" cargo build` on the `regex` crate:

collapsed
expanded

andjo403:
fixed the comment

PR #85, opened 2019-11-11
Make sure all crates compile on nightly and beta.

michaelwoerister:
Using the crate as part of `rustc` requires that it build on nightly. Let's make sure we catch cases early where it doesn't. This PR contains a fix for such a problem.

r? @wesleywiser

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
 language: rust
 rust:
   - stable
+  - beta
+  - nightly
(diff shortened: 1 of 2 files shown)


andjo403:
looks good.
good Idea to update the ci also.

andjo403:
with this I can remove the first commit in https://github.com/rust-lang/measureme/pull/83

PR #86, opened 2019-11-14
Only use 48 bits for encoding timestamps and 32 bits for encoding thread IDs in RawEvent in order to make it smaller.

michaelwoerister:
This restricts us to timestamps of at most ~78 hours after process start which seems acceptable. 
The PR reduces the size of `RawEvent` (and thus the size of our event stream files) by 25%.

### DIFF
--- a/measureme/src/file_header.rs
+++ b/measureme/src/file_header.rs
@@ -6,7 +6,7 @@ use crate::serialization::SerializationSink;
 use byteorder::{ByteOrder, LittleEndian};
 use std::error::Error;
 
-pub const CURRENT_FILE_FORMAT_VERSION: u32 = 1;
+pub const CURRENT_FILE_FORMAT_VERSION: u32 = 2;
 pub const FILE_MAGIC_EVENT_STREAM: &[u8; 4] = b"MMES";
 pub const FILE_MAGIC_STRINGTABLE_DATA: &[u8; 4] = b"MMSD";
 pub const FILE_MAGIC_STRINGTABLE_INDEX: &[u8; 4] = b"MMSI";
--- a/measureme/src/raw_event.rs
+++ b/measureme/src/raw_event.rs
@@ -1,30 +1,58 @@
 use crate::stringtable::StringId;
 
+/// `RawEvent` is how events are stored on-disk. If you change this struct,
+/// make sure that you increment `file_header::CURRENT_FILE_FORMAT_VERSION`.
 #[derive(Eq, PartialEq, Debug)]
 #[repr(C)]
 pub struct RawEvent {
     pub event_kind: StringId,
     pub event_id: StringId,
-    pub thread_id: u64,
-    pub start_ns: u64,
-    pub end_ns: u64,
+    pub thread_id: u32,
+
+    // The following 96 bits store the start and the end timestamp, using
+    // 48 bits for each.
+    pub start_time_lower: u32,
+    pub end_time_lower: u32,
+    pub start_and_end_upper: u32,
 }
+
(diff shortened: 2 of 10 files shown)


measureme/src/file_header.rs
```diff
@@ -6,7 +6,7 @@ use crate::serialization::SerializationSink;
 use byteorder::{ByteOrder, LittleEndian};
 use std::error::Error;
 
-pub const CURRENT_FILE_FORMAT_VERSION: u32 = 1;
+pub const CURRENT_FILE_FORMAT_VERSION: u32 = 2;
```
wesleywiser: I don't think we have adopted a formal policy yet, but since we haven't yet released format `1`, does this actually need to be incremented?

measureme/src/raw_event.rs
```diff
@@ -1,48 +1,100 @@
 use crate::stringtable::StringId;
 
+/// `RawEvent` is how events are stored on-disk. If you change this struct,
```
wesleywiser: I would love to see a test or static assertion that verifies the size of `RawEvent` is what we think it is.

measureme/src/file_header.rs
michaelwoerister: I think the policy should be that every PR keeps this consistent. It's not much work and there is no harm in jumping a few version numbers in between releases.

measureme/src/raw_event.rs
michaelwoerister: That's a good idea!

michaelwoerister:
Note to myself for tomorrow: Make event encoding endianess independent as part of this PR.

measureme/src/raw_event.rs
michaelwoerister: Done.

michaelwoerister:
I added two commits:
- Store events properly in little endian format on disk.
- Optimize the `FileSerializationSink` by directly copying data to the writing buffer.

The second commit makes the `FileSerializationSink` exactly as fast as the `MmapSerializationSink` in the benchmarks we have. But I only tested on Linux and I also remember that the benchmarks were no good indication of performance when used in `rustc`. So maybe that commit should be split out into a separate PR.

michaelwoerister:
I went ahead and move the last commit to a separate, less urgent PR (https://github.com/rust-lang/measureme/pull/88). This PR is ready for review.

measureme/src/raw_event.rs
```diff
@@ -1,46 +1,160 @@
...
+            // manually on big endian targets.
+            use byteorder::{LittleEndian, ByteOrder};
+
+            LittleEndian::write_u32(&mut bytes[0..], self.event_kind.as_u32());
+            LittleEndian::write_u32(&mut bytes[4..], self.event_id.as_u32());
+            LittleEndian::write_u32(&mut bytes[8..], self.thread_id);
+            LittleEndian::write_u32(&mut bytes[12..], self.start_time_lower);
+            LittleEndian::write_u32(&mut bytes[16..], self.end_time_lower);
+            LittleEndian::write_u32(&mut bytes[20..], self.start_and_end_upper);
+        }
```
eddyb: Out of curiosity, was the "manual" version benchmarked on x64?
The only reason I can think of is that it might be slower to write six `u32`s, one at a time (as opposed to 3 `u64` or one `u128` and one `u64`), but that's possible to alleviate, but first serializing to larger integers and then writing them out.

michaelwoerister: IIRC the two versions were almost the same with the manual one having one instruction more (and the whole fn being something 10 instructions). I don't expect there to be much difference in practice since there is more expensive stuff going (like writing this data to disk).

Issue #81, opened 2019-11-07
"summarize diff" could show relative statistic numbers


alexcrichton:
I was [looking at the output of `summarize diff`](https://gist.github.com/alexcrichton/03306558740b8f4363a18482e0653cc5) recently and was able to get some useful information (wow was LLVM a lot slower!). I think it'd be useful to also have information, though, for things like "how much slower" or "how much faster" are the two builds. For example it'd be cool to have an inline statistic saying like "LLVM is 100% slower" or "`is_copy_raw` is 10% slower". The absolute times are great but having relative times can also be good for putting things in context!

andjo403:
@alexcrichton is it an per item relative number that you think of or is it how mush this item have contributed to the total time change?
e.g. maybe not that interesting to see that is_copy_raw is 3ms slower when the build is 11s slower and LLVM_thin_lto_optimize is 3s slower.

alexcrichton:
Yeah I think per-item would be ideal, for example 100ms to 3100ms is crazy huge while 0.1ms to 3ms is miniscule

PR #87, opened 2019-11-14
add self_time_change to summarize diff

andjo403:
self_time_change is percentage change of self_time from base.
change to if base have value but the change do not
the base values is inverted and if the change have value
but the base do not the changes values is used.

fixes https://github.com/rust-lang/measureme/issues/81


andjo403:
this PR results in printout like 
image
image

PR #88, opened 2019-11-15
Optimize FileSerializationSink by using parking_lot::Mutex and avoiding heap allocations in write_atomic.

michaelwoerister:
This PR makes the `FileSerializationSink` exactly as fast as the `MmapSerializationSink` in the benchmarks we have. But I only tested on Linux and I also remember that the benchmarks were no good indication of performance when used in `rustc`.

It would be nice if we could get rid of the `MmapSerializationSink` because it keeps everything in memory until the end.

I wonder how much work it would be to have benchmarks that actually run `rustc`. It's a hassle to test this manually.

### DIFF
--- a/measureme/Cargo.toml
+++ b/measureme/Cargo.toml
@@ -15,6 +15,7 @@ travis-ci = { repository = "rust-lang/measureme" }
 [dependencies]
 byteorder = "1.2.7"
 rustc-hash = "1.0.1"
+parking_lot = "0.9"
 
 [target.'cfg(not(target_arch="wasm32"))'.dependencies]
 memmap = "0.6.0"
--- a/measureme/src/file_serialization_sink.rs
+++ b/measureme/src/file_serialization_sink.rs
@@ -1,12 +1,19 @@
 use crate::serialization::{Addr, SerializationSink};
 use std::error::Error;
 use std::fs;
-use std::io::{BufWriter, Write};
+use std::io::{Write};
 use std::path::Path;
-use std::sync::Mutex;
+use parking_lot::Mutex;
 
 pub struct FileSerializationSink {
-    data: Mutex<(BufWriter<fs::File>, u32)>,
+    data: Mutex<Inner>,
+}
+
+struct Inner {
+    file: fs::File,
+    buffer: Vec<u8>,
+    buf_pos: usize,
+    addr: u32,
 }
 
 impl SerializationSink for FileSerializationSink {
@@ -16,7 +23,12 @@ impl SerializationSink for FileSerializationSink {
         let file = fs::File::create(path)?;
 
         Ok(FileSerializationSink {
-            data: Mutex::new((BufWriter::new(file), 0)),
+            data: Mutex::new(Inner {
+                file,
+                buffer: vec![0; 1024*512],
+                buf_pos: 0,
+                addr: 0
+            }),
         })
     }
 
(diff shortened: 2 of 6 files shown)


michaelwoerister:
I think our benchmarks generate too little data to be realistic (less than a megabyte, I think). I'll look into that later. Also, a multi-threaded benchmark would be good.

wesleywiser:
> But I only tested on Linux and I also remember that the benchmarks were no good indication of performance when used in rustc.

FYI, as I recall, the issue was that mmap was quite a bit slower on Windows.

michaelwoerister:
Here are some benchmark numbers with and without the patch (numbers in milliseconds):

| Windows         |  before | after | improvement |
|:----------------|:-------:|:-----:|:-----------:|
| file, 1 thread  | 897     | 773   |    ~15%     |
| file, 8 threads | 921     | 663   |    ~30%     |
| mmap, 1 thread  | 909     | 939   |     -       |
| mmap, 8 threads | 839     | 830   |     -       |

| Linux           |  before | after | improvement |
|:----------------|:-------:|:-----:|:-----------:|
| file, 1 thread  | 605     | 510   |   15-20%    |
| file, 8 threads | 765     | 752   |      -      |
| mmap, 1 thread  | 512     | 519   |      -      |
| mmap, 8 threads | 378     | 349   |      -      |

The patch should be very good for Windows, especially for the multithreaded case. On Linux mmap is as fast (1 thread) or much faster (8 threads). I don't have any numbers for macOS.

wesleywiser:
I can gather some macOS numbers if you're interested.

michaelwoerister:
If you have time, I'd be interested, yeah. 

I got the numbers by running `cargo +nightly bench`. The "after" numbers are from this PR as it is, and the "before" numbers are from this PR, but with commit 8367ec654b31ed73c14f50addbc76388aa261f6b removed. The last commit has to stay in because it modifies the benchmarks.

michaelwoerister:
Update: I just pushed the "baseline" version to https://github.com/michaelwoerister/measureme/tree/opt-file-sink-ref.

wesleywiser:
Here's what I'm seeing on my 4 core 8 thread MBP:

| macOS         |  before | after | improvement |
|:----------------|:-------:|:-----:|:-----------:|
| file, 1 thread  | 759     |  556  |    ~26%     |
| file, 8 thread | 1,666 | 559 | ~66% |
| mmap, 1 thread | 670 | 622 | ~7% |
| mmap, 8 thread | 394 |386 | ~2% |

michaelwoerister:
Thanks, @wesleywiser! Looks like we actually might want to switch to `FileSerializationSink` on macOS too for the non-parallel case.

measureme/src/file_serialization_sink.rs
```diff
@@ -25,17 +37,45 @@ impl SerializationSink for FileSerializationSink {
...
+            ref mut file,
+            ref mut buffer,
+            ref mut buf_pos,
+            ref mut addr
+        } = *data;
 
-        let mut data = self.data.lock().expect("couldn't acquire lock");
-        let curr_addr = data.1;
-        let file = &mut data.0;
+        assert!(num_bytes <= buffer.len());
```
wesleywiser: If `num_bytes > buffer.len()` we could flush the buffer and then write to the file directly.

measureme/Cargo.toml
```diff
@@ -15,6 +15,7 @@ travis-ci = { repository = "rust-lang/measureme" }
 [dependencies]
 byteorder = "1.2.7"
 rustc-hash = "1.0.1"
+parking_lot = "0.9"
```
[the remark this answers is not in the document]
michaelwoerister: Yeah, I think performance on Windows is much better with parking_lot but let me verify that. I do like to keep this library lean.

measureme/src/file_serialization_sink.rs
michaelwoerister: I'll look into that.

michaelwoerister:
The second commit is still work-in-progress, yes. I want to update the code in testing_common to handle multiple threads, which shouldn't be too hard. Thanks for review!

michaelwoerister:
I did some new Windows measurements of `parking_lot` vs `std::sync`. It looks like `parking_lot` is a bit faster indeed. Let's keep it.

| Windows         |  before | std::sync | parking_lot |
|:----------------|:-------:|:-----:|:-----------:|
| file, 1 thread  | 744     | 614   |   589     |
| file, 8 threads | 862     | 637   |    582     |

measureme/src/file_serialization_sink.rs
michaelwoerister: Implemented in https://github.com/rust-lang/measureme/pull/88/commits/b9d011168e664cd1cdf15b4f24838c4821ac6f9f.

michaelwoerister:
This is ready for review now.

measureme/src/file_serialization_sink.rs
```diff
@@ -16,7 +23,12 @@ impl SerializationSink for FileSerializationSink {
         let file = fs::File::create(path)?;
 
         Ok(FileSerializationSink {
-            data: Mutex::new((BufWriter::new(file), 0)),
```
andjo403: what is the gain by removing the BufWriter? feels like the new code is similar to the BufWriter code
so will BufWriter::with_capacity(1024*512, file)  give the same result?

andjo403:
have someone looked at how mush the result differ for the "fast path" where the buffer is updated and the "slow path" where the file is written? as we are blocking all threads from writing during the file write this can affect many events.
also by increasing the size of the buffer to 512Kb from 8Kb maybe the variance of the measurements bigger as it takes longer to write.

measureme/src/file_serialization_sink.rs
michaelwoerister: The difference is that `BufWriter` does not allow directly writing to its buffer, so we are basically re-implementing `BufWriter` here. The interface of `write_atomic` requires there to be a writable output buffer.

michaelwoerister:
@andjo403 I have not investigated variance. The bigger buffer should reduce the number of writes, while making each write larger. So they are less evenly distributed but the fixed overhead might amortize better.

It would be nice to do the actual file writing in a background thread via some kind of double buffering scheme. I haven't tried to implement something like that though.

PR #89, opened 2019-11-16
Add LightweightEvent for making `Event` string fields in lazy

### LINKED ISSUE #80: Consider making string fields in `Event` lazy
There's quite a bit of post-processing code that iterates over events without looking at strings. It might make sense to lazily read strings from the string table.

silathdiir:
Tries to fix #80 according to @michaelwoerister 's comments.

### DIFF
--- a/analyzeme/src/lightweight_event.rs
+++ b/analyzeme/src/lightweight_event.rs
@@ -1,0 +1,85 @@
+use crate::event::Event;
+use crate::profiling_data::ProfilingData;
+use crate::timestamp::Timestamp;
+use std::hash::{Hash, Hasher};
+use std::time::Duration;
+
+#[derive(Clone, Debug)]
+pub struct LightweightEvent<'a> {
+    pub data: &'a ProfilingData,
+    pub event_index: usize,
+    pub thread_id: u32,
+    pub timestamp: Timestamp,
+}
+
+impl<'a> LightweightEvent<'a> {
+    pub fn to_event(&self) -> Event<'a> {
+        self.data.decode_full_event(self.event_index)
+    }
+
+    /// Returns true if the time interval of `self` completely contains the
+    /// time interval of `other`.
+    pub fn contains(&self, other: &LightweightEvent) -> bool {
+        match self.timestamp {
+            Timestamp::Interval {
+                start: self_start,
+                end: self_end,
+            } => match other.timestamp {
+                Timestamp::Interval {
+                    start: other_start,
+                    end: other_end,
+                } => self_start <= other_start && other_end <= self_end,
+                Timestamp::Instant(other_t) => self_start <= other_t && other_t <= self_end,
+            },
+            Timestamp::Instant(_) => false,
+        }
+    }
+
+    pub fn duration(&self) -> Option<Duration> {
+        match self.timestamp {
+            Timestamp::Interval { start, end } => end.duration_since(start).ok(),
+            Timestamp::Instant(_) => None,
+        }
+    }
+}
+
+impl<'a> PartialEq for LightweightEvent<'a> {
+    fn eq(&self, other: &LightweightEvent<'a>) -> bool {
+        let LightweightEvent {
+            data,
+            event_index,
+            thread_id,
+            timestamp,
+        } = *self;
+
+        let LightweightEvent {
+            data: other_data,
+            event_index: other_event_index,
+            thread_id: other_thread_id,
+            timestamp: other_timestamp,
+        } = *other;
+
+        std::ptr::eq(data, other_data) &&
+            event_index == other_event_index &&
+            thread_id == other_thread_id &&
+            timestamp == other_timestamp
+    }
+}
+
+impl<'a> Eq for LightweightEvent<'a> {}
+
+impl<'a> Hash for LightweightEvent<'a> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let LightweightEvent {
+            data,
+            event_index,
+            thread_id,
+            timestamp,
+        } = *self;
+
+        std::ptr::hash(data, state);
+        event_index.hash(state);
+        thread_id.hash(state);
+        timestamp.hash(state);
+    }
+}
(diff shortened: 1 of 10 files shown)


andjo403:
can be good to change the wording of the closing comment above to auto close the issue at merge see [closing issues using keywords](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords)

silathdiir:
@andjo403 Thanks, I updated the description.

analyzeme/src/lightweight_event.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+impl<'a> Eq for LightweightEvent<'a> {}
+
+impl<'a> Hash for LightweightEvent<'a> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        std::ptr::hash(self.data, state);
+        self.event_index.hash(state);
+        self.timestamp.hash(state);
+        self.thread_id.hash(state);
```
michaelwoerister: While doing lots of `HashStable` impls in the compiler I've really come to appreciate using this "exhaustive destructuring" pattern in order not to forget any fields (and guard against forgetting newly added fields): 
```rust
let LightweightEvent {
    data,
    event_index,
    timestamp,
    thread_id,
} = *self;

std::ptr::hash(data, state);
event_index.hash(state);
timestamp.hash(state);
thread_id.hash(state);
```

It's a bit more verbose but it enables the compiler to help you with a common error case.

analyzeme/src/profiling_data.rs
```diff
@@ -78,33 +81,17 @@ impl ProfilingData {
...
-    pub fn new(data: &'a ProfilingData) -> ProfilerEventIterator<'a> {
-        ProfilerEventIterator {
-            data,
-            forward_event_idx: 0,
-            backward_event_idx: data.num_events(),
-        }
-    }
 
-    fn decode_event(&self, event_index: usize) -> Event<'a> {
+    pub fn decode_full_event<'a>(&'a self, event_index: usize) -> Event<'a> {
```
michaelwoerister: I'd like to keep this `pub(crate)` for now as I don't want to guarantee fast random access to events in the profiling data. That would make it harder for us to ever switch to a variable length encoding or having a single file for events + stringtable.

analyzeme/src/profiling_data.rs
```diff
@@ -114,17 +101,50 @@ impl<'a> ProfilerEventIterator<'a> {
             thread_id: raw_event.thread_id,
         }
     }
+
+    pub fn decode_lightweight_event<'a>(&'a self, event_index: usize) -> LightweightEvent<'a> {
```
michaelwoerister: Same here with `pub(crate)`.

summarize/src/analysis.rs
```diff
@@ -110,7 +110,7 @@ use std::time::SystemTime;
 /// `e4` and `e3` need to be popped in the same step.
 pub fn perform_analysis(data: ProfilingData) -> Results {
     struct PerThreadState<'a> {
-        stack: Vec<Event<'a>>,
+        stack: Vec<LightweightEvent<'a>>,
```
michaelwoerister: Since we basically always expand events to full events here, I think, it's better to keep the stack a `Vec<Event<'a>>` and just change the `for` loop below to `for current_event in data.iter().rev().map(LightweightEvent::to_event) { ... }`.

analyzeme/src/lightweight_event.rs
michaelwoerister: This comment also applies to the `PartialEq` impl above.

analyzeme/src/profiling_data.rs
silathdiir: Hi @michaelwoerister , for keeping this function private, I tries to add a function pointer `event_decoder` to LightweightEvent.

I could not confirm if it is the right solution. Please help review the update code. Thanks a lot.

analyzeme/src/lightweight_event.rs
```diff
@@ -7,14 +7,15 @@ use std::time::Duration;
...
+    pub event_decoder: fn(&'a ProfilingData, usize) -> Event<'a>,
     pub event_index: usize,
-    pub timestamp: Timestamp,
     pub thread_id: u32,
+    pub timestamp: Timestamp,
 }
 
 impl<'a> LightweightEvent<'a> {
     pub fn to_event(&self) -> Event<'a> {
-      self.data.decode_full_event(self.event_index)
```
michaelwoerister: You can keep this line as you wrote it originally.

analyzeme/src/lightweight_event.rs
```diff
@@ -7,14 +7,15 @@ use std::time::Duration;
 #[derive(Clone, Debug)]
 pub struct LightweightEvent<'a> {
     pub data: &'a ProfilingData,
+    pub event_decoder: fn(&'a ProfilingData, usize) -> Event<'a>,
```
michaelwoerister: This won't be needed.

analyzeme/src/profiling_data.rs
```diff
@@ -82,7 +82,7 @@ impl ProfilingData {
         event_byte_count / RAW_EVENT_SIZE
     }
 
-    pub fn decode_full_event<'a>(&'a self, event_index: usize) -> Event<'a> {
+    fn decode_full_event<'a>(&'a self, event_index: usize) -> Event<'a> {
```
michaelwoerister: Write this line as `pub(crate) fn decode_full_event<'a>(&'a self, event_index: usize) -> Event<'a> {`, then `decode_full_event` can be used from within the entire `analyzeme` crate, but not from outside of the crate. That's a sensible level of visibility for our purposes.

analyzeme/src/profiling_data.rs
```diff
@@ -298,6 +299,10 @@ impl ProfilingDataBuilder {
 
 impl<'a> ExactSizeIterator for ProfilerEventIterator<'a> {}
 
+fn decode_event<'a>(data: &'a ProfilingData, event_index: usize) -> Event<'a> {
+    data.decode_full_event(event_index)
+}
+
```
michaelwoerister: This won't be needed.

analyzeme/src/profiling_data.rs
silathdiir: Thanks a lot. Will update the code.

PR #91, opened 2019-12-01
measureme: Update memmap to 0.7.0, and bump version to 0.4.1

jackpot51:
memmap 0.7.0 has support for Redox, hence the need for this change. No compatibility issues required fixing, it still passes all tests.

### DIFF
--- a/measureme/Cargo.toml
+++ b/measureme/Cargo.toml
@@ -17,4 +17,4 @@ byteorder = "1.2.7"
 rustc-hash = "1.0.1"
 
 [target.'cfg(not(target_arch="wasm32"))'.dependencies]
-memmap = "0.6.0"
+memmap = "0.7.0"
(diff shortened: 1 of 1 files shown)


jackpot51:
There may be other changes since 0.4.0 that necessitate a minor version bump. Please let me know if you want me to bump it to 0.5.0 instead.

wesleywiser:
Thanks for the PR! We've been making other breaking changes since the 0.4 release so the next release will be 0.5. I'd suggest for now, removing the second commit which bumps the measurme version. We'll take care of that when we make the release.

wesleywiser:
Just out of curiosity, any particular reason we're going to `memmap@0.7`? rustc is still on `0.6.2`.

michaelwoerister:
I'll include this change in my version bump to 0.5. Thanks, @jackpot51!

PR #92, opened 2019-12-02
Bump version to 0.5

michaelwoerister:
r? @wesleywiser 

cc @jackpot51

### DIFF
--- a/measureme/Cargo.toml
+++ b/measureme/Cargo.toml
@@ -18,4 +18,4 @@ rustc-hash = "1.0.1"
 parking_lot = "0.9"
 
 [target.'cfg(not(target_arch="wasm32"))'.dependencies]
-memmap = "0.6.0"
+memmap = "0.7"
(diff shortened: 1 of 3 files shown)


michaelwoerister:
I'd like to start using this version in `rustc` as soon as possible.

measureme/Cargo.toml
```diff
@@ -18,4 +18,4 @@ rustc-hash = "1.0.1"
 parking_lot = "0.9"
 
 [target.'cfg(not(target_arch="wasm32"))'.dependencies]
-memmap = "0.6.0"
+memmap = "0.7"
```
wesleywiser: Do we also need to handle bumping `memmap` upstream in rustc? Or is it ok to have two different versions of `memmap` compiled in?

michaelwoerister: I think it's not a problem to have different versions (except for bloat), but I plan to also bump `memmap` on the `rustc` side in the same PR that bumps `rustc`'s `measureme` version.

michaelwoerister:
Thanks for the quick review, @wesleywiser!

Issue #75, opened 2019-11-06
Event recording & summarize need cleanup with respect to self-time vs incr-load-time vs blocked-time vs total-time.


michaelwoerister:
At the moment, we have not specified how self-time and other kinds of measurements interact with each other:

- Does the incremental cache loading time contribute to a queries self-time?
- In the incremental cache loading case, is the loading event nested inside a regular query-provider event?
- Should we distinguish between actual cache loading, metadata loading, and provider re-executions caused by missing cache entries? If so, how?
- Do query-blocked time contribute to self-time? How about total-time?
- Is a query-blocked event always followed by a cache-hit event (the latter of which would then be responsible for incrementing the invocation count) What if cache-hit events are filtered out? Is it OK to have wrong invocation counts?

michaelwoerister:
My suggested answers to these questions are:

> Does the incremental cache loading time contribute to a queries self-time?

Yes. Self-time is what one sorts for when trying to find out what is expensive. It would be unhelpful if something got sorted to the back even though 50% of total compile time is spent loading things for it from the cache.

> In the incremental cache loading case, is the loading event nested inside a regular query-provider event?

Let's go with no. That causes fewer events to be generated. Also, in order to make things really uniform, in-memory cache hit events would have to wrapped with a query-provider event too, which would be lots of overhead for not much gain. 

> Should we distinguish between actual cache loading, metadata loading, and provider re-executions caused by missing cache entries? If so, how?

Down the line, yes. It's a useful distinction to make. These should all be separate query kinds (incr. cache loading already is). However, I think this is something to be implemented after we have query keys, so we don't make too many changes at once.

> Do query-blocked time contribute to self-time? How about total-time?

Blocking should add to self-time for the same reason as incr. cache loading should add to self-time (see above). I don't remember what I meant by "How about total-time?", I think the question is not relevant anymore since we switched to @andjo403's approach of computing total time as `thread_max_time - thread_min_time`. 

> Is a query-blocked event always followed by a cache-hit event (the latter of which would then be responsible for incrementing the invocation count) What if cache-hit events are filtered out? Is it OK to have wrong invocation counts?

I haven't thought about this one yet.

michaelwoerister:
> Is a query-blocked event always followed by a cache-hit event (the latter of which would then be responsible for incrementing the invocation count) What if cache-hit events are filtered out? Is it OK to have wrong invocation counts?

[...]

PR #93, opened 2019-12-03
Support query events stacked on incremental_load events in summarize.

michaelwoerister:
Fixes the issue found in https://github.com/rust-lang/rust/pull/66981. The should be correct but this could be cleaned up even more if the questions in https://github.com/rust-lang/measureme/issues/75 were properly answered. Would be good to do that before the next bigger release.

r? @wesleywiser or @andjo403

### DIFF
--- a/summarize/src/analysis.rs
+++ b/summarize/src/analysis.rs
@@ -159,11 +159,25 @@ pub fn perform_analysis(data: ProfilingData) -> Results {
                     thread.stack.pop();
                 }
 
+                let current_event_duration = current_event.duration().unwrap();
+
                 // If there is something on the stack, subtract the current
                 // interval from it.
                 if let Some(current_top) = thread.stack.last() {
                     record_event_data(&current_top.label, &|data| {
-                        data.self_time -= current_event.duration().unwrap();
+                        match &current_top.event_kind[..] {
+                            QUERY_EVENT_KIND | GENERIC_ACTIVITY_EVENT_KIND => {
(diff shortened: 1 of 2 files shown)


summarize/src/analysis.rs
```diff
@@ -159,33 +159,47 @@ pub fn perform_analysis(data: ProfilingData) -> Results {
...
                     record_event_data(&current_top.label, &|data| {
-                        data.self_time -= current_event.duration().unwrap();
+                        match &current_top.event_kind[..] {
+                            QUERY_EVENT_KIND | GENERIC_ACTIVITY_EVENT_KIND => {
+                                data.self_time -= current_event_duration;
+                            }
+                            INCREMENTAL_LOAD_RESULT_EVENT_KIND => {
+                                data.incremental_load_time  -= current_event_duration;
+                            }
+                            _ => {
```
wesleywiser: We probably need to handle `QUERY_BLOCKED_EVENT_KIND` here.

michaelwoerister: Hm, that definitely shouldn't have anything on top of it, right?

wesleywiser: I'm not sure I know enough about how the query system works when parallelism is turned on to say. Does the thread stay blocked when a query is blocked or does it try to work on another query?

michaelwoerister: It stays blocked.

wesleywiser: Ah ok. You're right then.

Issue #96, opened 2019-12-09
Consider getting rid of string index data file


michaelwoerister:
The new file encoding for the string table introduced in https://github.com/rust-lang/measureme/pull/90 still has some unused bit-prefixes in its grammar that would allow it to efficiently (i.e. without additional space overhead) encode the string index data table as part of the regular string table data. The index entries would just be interspersed with the string components and the grammar would make sure that we can parse them out properly again. 

The only reason I'm a bit hesitant to implement this is that it would require tools to always parse the entire string table once on startup in order to find all the index entries. It's just a single pass and the parsing is not more complicated than parsing a huge UTF-8 string, but it would add a bit of overhead to every tool invocation. 

On the positive side it would allow us to get rid of one of the three files we generate when recording a profile.

Mark-Simulacrum:
I am in favor of reducing the number of files. I suspect the cost of parsing the huge UTF-8 string is pretty low? IIRC, UTF-8 decoding is on-par with disk read speed (or faster) these days.

michaelwoerister:
I suspect that the slowdown wouldn't be noticeable but I we don't have benchmarks for the tools, I think.

PR #98, opened 2019-12-09
Virtual instead of reserved string ids

michaelwoerister:
This PR removes some unnecessary overhead from the StringTable format. 

> With this commit only "virtual" StringIds get an entry in the index table and regular StringIds
store an actual address instead of index table key. That makes the index data a lot smaller
and removes the need to do a table lookup for regular StringIds.

This optimization becomes important when recording query keys because we allocate a lot more string data in that case.

~~This PR also contains the changes from https://github.com/rust-lang/measureme/pull/97. Feel free to review/merge all at once or separately.~~

### DIFF
--- a/analyzeme/src/stringtable.rs
+++ b/analyzeme/src/stringtable.rs
@@ -12,13 +12,9 @@ use std::borrow::Cow;
 use std::error::Error;
 use memchr::memchr;
 
-// See module-level documentation for more information on the encoding.
-const UTF8_CONTINUATION_MASK: u8 = 0b1100_0000;
-const UTF8_CONTINUATION_BYTE: u8 = 0b1000_0000;
-
 fn deserialize_index_entry(bytes: &[u8]) -> (StringId, Addr) {
     (
-        StringId::reserved(LittleEndian::read_u32(&bytes[0..4])),
+        StringId::new(LittleEndian::read_u32(&bytes[0..4])),
         Addr(LittleEndian::read_u32(&bytes[4..8])),
     )
 }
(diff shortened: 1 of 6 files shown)


measureme/src/stringtable.rs
```diff
@@ -233,41 +267,63 @@ impl<S: SerializationSink> StringTableBuilder<S> {
...
+                let id = from.0;
+                assert!(id <= MAX_USER_VIRTUAL_STRING_ID);
+                (id.to_le(), to_addr_le)
+            })
+            .collect();
+
+        let num_bytes = serialized.len() * std::mem::size_of::<MappingEntry>();
+        let byte_ptr = serialized.as_ptr() as *const u8;
+
+        let bytes = unsafe { std::slice::from_raw_parts(byte_ptr, num_bytes) };
```
wesleywiser: According to [this](https://doc.rust-lang.org/reference/type-layout.html#tuple-layout) tuples don't have a defined layout so I don't think this is technically allowed? Switching to a `[u32; 2]` array should be ok.

analyzeme/src/stringtable.rs
```diff
@@ -87,6 +117,32 @@ impl<'st> StringRef<'st> {
...
+
+fn is_utf8_continuation_byte(byte: u8) -> bool {
+    // See module-level documentation for more information on the encoding.
+    const UTF8_CONTINUATION_MASK: u8 = 0b1100_0000;
+    const UTF8_CONTINUATION_BYTE: u8 = 0b1000_0000;
+    (byte & UTF8_CONTINUATION_MASK) == UTF8_CONTINUATION_BYTE
+}
+
+// String IDs in the table data are encoded in big endian format, while string
+// IDs in the index are encoded in little endian format. Don't mix the two up.
```
wesleywiser: Why do we use two different formats? Is it because of the utf-8 stuff we're doing?

measureme/src/stringtable.rs
michaelwoerister: Oh! Great catch!

analyzeme/src/stringtable.rs
michaelwoerister: Yes, we need big endian so that the uppermost bits end up in the first byte we encounter. But for the index table that's not necessary, so little endian is the more efficient option (at least on little endian platforms, which means all the platforms `:)`). We _probably_ could use big endian everywhere without noticing at real performance difference though.

michaelwoerister: If we implemented https://github.com/rust-lang/measureme/issues/96 then this issue would solve itself too.

measureme/src/stringtable.rs
michaelwoerister: Updated.

PR #103, opened 2020-01-02
Fix big-endian build and add CI tests for it.

michaelwoerister:
This PR

- fixes a compilation error on big endian platforms
- adds a big endian `cargo check` build to CI, so we catch compilation errors like these early next time
- bumps the version to `0.7.1` so we can use that to unblock https://github.com/rust-lang/rust/pull/67397.

~~This PR should fail until the big endian specific error uncovered [here](https://github.com/rust-lang/rust/pull/67397#issuecomment-568295509) is fixed.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,15 @@
 language: rust
+before_install:
+  - rustup target add powerpc64-unknown-linux-gnu
 rust:
   - stable
   - beta
   - nightly
+fast_finish: true
+env:
+  - CARGO_INCREMENTAL=0
+install: skip
+script:
+  - cargo check --verbose --target powerpc64-unknown-linux-gnu --lib --bins --tests || exit 1
+  - cargo build --verbose --all || exit 1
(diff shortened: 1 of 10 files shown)


michaelwoerister:
But it fails before it can even install Rust `:(`

wesleywiser:
https://github.com/rust-lang/measureme/blob/560162a2e81c3efec7c89837dc7ed78105450a4c/measureme/src/raw_event.rs#L152-L153

is `cfg`'d out on little-endian and got broken in a prior commit. I believe these should just be `StringId::new()` right?

michaelwoerister:
Yes, that's the error. I'm tweaking `.travis.yml` until it properly catches the error and then add a fix to this PR.

wesleywiser:
Oh, sorry. I didn't realize what was going on. Carry on! 🙂

michaelwoerister:
Alright, ready for review `:)`

Issue #67, opened 2019-10-16
summarize could show min/max times and other useful stats.


eddyb:
We're investigating a case where a query is taking 30-40ms on average, but by using `crox` we can see instances of that query being over 200ms sometimes, with the maximum around 330ms.

Would be nice to get this information from `summarize` itself.

theotherphil:
I'd like to have a go at this, if it's not particularly time sensitive.

michaelwoerister:
A related question to @Mark-Simulacrum: Will perf.rlo crash if new columns are added or will it just ignore them?

michaelwoerister:
@theotherphil Feel free to give it a try!

Mark-Simulacrum:
I'm not sure what you mean by columns. I expect no - perf.rlo parses JSON summaries with serde and iirc doesn't prohibit unknown fields and such, though it may come down to the specifics of the data format. I don't expect problems though.

michaelwoerister:
With "column" I mean something to would end up as a new column in the detailed view table, so in practice a new field in `summarize`'s [QueryData struct](https://github.com/rust-lang/measureme/blob/e6f55c7bcacc1c70b895c858d1840958709b9a34/summarize/src/query_data.rs#L7).

I gave it a try earlier and, as you say, `serde_json` seems to handle unknown fields just fine it seems: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4c8b1574ce2e46c4e4687d88e7938c98

theotherphil:
I've just had a look at this.

I want to start by adding min and max, to get the wiring sorted before I worry about the overhead of tracking the state required to compute percentiles.

I'm a bit stuck at the very first hurdle: I'm not sure how to recognise a complete logical event in the event data. Specifically, I'm not sure what relationship is guaranteed between QUERY_EVENT_KIND, INCREMENTAL_LOAD_RESULT_EVENT_KIND and QUERY_BLOCKED_EVENT_KIND events in rustc's self-profiler output. https://github.com/rust-lang/measureme/pull/104 adds the time from incremental cache loading and blocked queries to the self time for each event label. Should I attempt to group consecutive events with different kinds but matching labels, or just ignore the latter two event kinds?

michaelwoerister:
@theotherphil Each event is self-contained. For min/max you can just compare `Event::timestamp`s. I would not make a distinction between the different event kinds.

PR #104, opened 2020-01-03
summarize: Make incr-cache-loading and query-blocking contribute to self-time of query.

michaelwoerister:
Adjusts the logic in `summarize` according to https://github.com/rust-lang/measureme/issues/75#issuecomment-563137947. 

The differences in absolute numbers should be small. But it is more consistent, I think.

Fixes https://github.com/rust-lang/measureme/issues/75.

r? @wesleywiser

### DIFF
--- a/summarize/src/analysis.rs
+++ b/summarize/src/analysis.rs
@@ -175,6 +175,7 @@ pub fn perform_analysis(data: ProfilingData) -> Results {
                                 data.self_time -= current_event_duration;
                             }
                             INCREMENTAL_LOAD_RESULT_EVENT_KIND => {
+                                data.self_time -= current_event_duration;
                                 data.incremental_load_time -= current_event_duration;
                             }
                             _ => {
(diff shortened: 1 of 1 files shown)


michaelwoerister:
I might make sense to do a new release for this and switch perf.rlo over. It is the last thing missing for the MVP, I'd say.

andjo403:
to me this looks good.
but maybe before a new release, something that you talked about in the issue was total time and maybe you are thinking the total time for a label the same as in chrome tracing where you have a self time and total time for an event and I think that is missing from showing all information for an events. I also think that the total time is needed for e.g. in https://github.com/rust-lang/measureme/issues/67

michaelwoerister:
@andjo403, I think what you are talking about I have seen being referred to as "inclusive" and "exclusive" times for an event. "Inclusive" would then also contain the time of child events. I think calling it "inclusive" and "exclusive" times is pretty good. Right now we only show "exclusive" times (what we call "self time"), adding a column for inclusive time to perf.rlo is a very good idea.

michaelwoerister:
Hm, kCacheGrind and https://github.com/KDAB/hotspot call it "self" and "inclusive" times.

wesleywiser:
The Chrome profile calls this "self time" vs "total time" FWIW

PR #108, opened 2020-01-13
add additinonal data to crox events

andjo403:
hade been nice to have some reserved names (eg. def_id, cgu_id ...) for the additional data now I have only enumerated them.

eg. of how it looks in chrome://tracing/
image

### DIFF
--- a/crox/src/main.rs
+++ b/crox/src/main.rs
@@ -116,6 +116,21 @@ fn generate_thread_to_collapsed_thread_mapping(
     thread_to_collapsed_thread
 }
 
+fn get_args(full_event: &analyzeme::Event) -> Option<FxHashMap<String, String>> {
+    if !full_event.additional_data.is_empty() {
+        Some(
+            full_event
+                .additional_data
+                .iter()
+                .enumerate()
+                .map(|(i, arg)| (format!("arg{}", i).to_string(), arg.to_string()))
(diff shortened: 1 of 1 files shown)


crox/src/main.rs
```diff
@@ -4,7 +4,7 @@ use std::io::BufWriter;
 use std::path::PathBuf;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-use analyzeme::{ProfilingData, Timestamp};
+use analyzeme::{ProfilingData, Timestamp, Event as analyzemeEvent};
```
wesleywiser: It's kind of weird to have identifiers that start with lower case letters in Rust. Why don't we just do

```suggestion
use analyzeme::{self, ProfilingData, Timestamp};
```

and then

```rust
fn get_args(full_event: &analyzeme::Event) -> Option<FxHashMap<String, String>> {
```

below?

andjo403: done

andjo403:
@michaelwoerister how do you feel about having the seperators inform about the type of argument that will come e.g. so def_id can be printed insted of arg0

michaelwoerister:
I think the main problem is that we don't have the parameter name available at the time of recording. In many cases we don't even define a name for the argument (and most of the time it's just `key`).

PR #109, opened 2020-01-13
add event time column to summarize results

### LINKED ISSUE #107: Add "incl. time" column to summarize and perf.rlo
Many tools show both the "self-time" of an event and the time that also includes the child events. I found myself wanting that feature for perf.rlo in a number of cases recently. Others seems to think the same: https://github.com/rust-lang/measureme/pull/104#issuecomment-571772190.

I think I would prefer the caption "incl. time" for this since we already use the term "total time" for the total amount of time spent in all threads.

andjo403:
name it only time for now feel that "incl. time" need more description then only having time and self-time. but feel that the "% of total time" get unclear now that we have time and self-time

closes #107


michaelwoerister:
Nice, thanks @andjo403!

One thing I'm wondering? How does this handle recursive invocations like:
```
       <---A--->
   <------B------>
<---------A--------->
```
How _should_ it handle cases like this? Always only count the bottom-most stack element?

andjo403:
as it is now it will count the time for both events in the example see the test https://github.com/rust-lang/measureme/pull/109/files#diff-98cefccab6f8905ece90aaf3c2cb1ea4R479 .
I think that is correct at least that is what I wanted the last time when I want to know how mush time is spent in this event. and feel that it will be strange if the result differs agains it was 3 diffrent events in the stack.

michaelwoerister:
I think an argument can be made for either way of counting. I think I'm fine with the way it's done here though.

Issue #114, opened 2020-06-01
summarize doesn't work if you specify your own event kinds


jasonwilliams:
In Boa we are working with measureme to measure performance, our main use case is the Chrome Profiler, but we were interested in using `summarize` also.

The solution seems to be to use the [generic eventKind](https://docs.rs/measureme/0.7.1/measureme/rustc/constant.GENERIC_ACTIVITY_EVENT_KIND.html) but that would mean having to remove our own kinds and on Chrome it would all be one colour, this would not be ideal for us.

Our output when using:
```
Ignoring event with unknown event kind `value`
Unexpectedly enountered event `Event { event_kind: "value", label: "set_field", additional_data: [], timestamp: Interval { start: SystemTime { intervals: 132353512591167719 }, end: SystemTime { intervals: 132353512591168006 } }, thread_id: 1 }`, while top of stack was `Event { event_kind: "init", label: "make_builtin_fn: toString", additional_data: [], timestamp: Interval { start: SystemTime { intervals: 132353512591167595 }, end: SystemTime { intervals: 132353512591168321 } }, thread_id: 1 }`. Ignoring.
Ignoring event with unknown event kind `value`
Unexpectedly enountered event `Event { event_kind: "object", label: "Object::set", additional_data: [], timestamp: Interval { start: SystemTime { intervals: 132353512591167802 }, end: SystemTime { intervals: 132353512591167977 } }, thread_id: 1 }`, while top of stack was `Event { event_kind: "value", label: "set_field", additional_data: [], timestamp: Interval { start: SystemTime { intervals: 132353512591167719 }, end: SystemTime { intervals: 132353512591168006 } }, thread_id: 1 }`. Ignoring.
Ignoring event with unknown event kind `object`
Unexpectedly enountered event `Event { event_kind: "object", label: "Object::get_internal_slot", additional_data: [], timestamp: Interval { start: SystemTime { intervals: 132353512591167920 }, end: SystemTime { intervals: 132353512591167934 } }, thread_id: 1 }`, while top of stack was `Event { event_kind: "object", label: "Object::set", additional_data: [], timestamp: Interval { start: SystemTime { intervals: 132353512591167802 }, end: SystemTime { intervals: 132353512591167977 } }, thread_id: 1 }`. Ignoring.
Ignoring event with unknown event kind `object`
[...]

Issue #112, opened 2020-02-24
`summarize` shouldn't show columns that don't apply to the profile


wesleywiser:
For example, most people do not enable query cache hit events or use a parallel compiler so the "Cache Hits" and "Blocked time" columns never have values. We should hide those columns in this case.

Mark-Simulacrum:
Please make sure (or let me know if otherwise) that changes here either do not affect the JSON output (i.e., we should have default values instead of omitting keys.

OTOH, if we do change the JSON output, then that's fine, I can handle it on perf's side, but a heads up would be appreciated.

PR #116, opened 2020-06-24
Make summarize useful to data producers other than rustc

wesleywiser:
The [boa project](https://github.com/boa-dev/boa) is using measureme to trace their own performance. This lets them use the `summarize` tool which was previously only useful to rustc because it cared about very specific event kinds.


andjo403:
nice that you remove the columns that do not have any data makes the printout cleaner

andjo403:
think that this also closes #114

wesleywiser:
Oh, good call. Thanks @andjo403!

andjo403:
hmm #112 also seems to be fixed missed that before

PR #118, opened 2020-07-11
Adjust top-level documentation for current state

nagisa:
The documentation had a number of out-dated references and broken links.
The links are replaced with intra-doc links and a lint is denied to
ensure they remain valid.

(Note, however that the intra-doc links only work with a nightly rustdoc AFAIK…)


wesleywiser:
This, is great! Thanks

PR #120, opened 2020-07-12
Add a tracing subscriber implementation

nagisa:


wesleywiser:
Sorry for the late review @nagisa!

nagisa:
I am not going to pursue this any further. While the idea sounds great on the surface, the differences in operation are big enough that it does not make too much sense to actually pursue something like this.

In particular, in `tracing` events are fairly important, and for the subscriber to be useful `measureme` would need to actively process them, however it does not currently because `rustc` emits so many of them that `measureme` does explicitly not process them at all.

wesleywiser:
Thanks for the analysis @nagisa! Your other changes look good so I'm going to pull those in.

PR #121, opened 2020-07-18
Add named "additional data" support

nagisa:
This allows representing KV pairs of data in e.g. chrome profiles more
faithfully. Previously we would automatically generate arg0... as a key
for all such data.

Chromium Profiler showing proper key-value information for the selected span


nagisa:
Note: this doesn’t actually add the API to measureme to output such data. Ideas on what API would make most sense here would be appreciated.

measureme/src/event_id.rs
```diff
@@ -1,20 +1,24 @@
...
-/// Event IDs are strings conforming to the following grammar:
+/// Event IDs are strings conforming to the following (eBNF) grammar:
 ///
 /// ```ignore
 ///   <event_id> = <label> {<argument>}
 ///   <label> = <text>
-///   <argument> = '\x1E' <text>
+///   <argument> = ['\x1D' <argument_name>] '\x1E' <argument_value>
+///   <argument_name> = <text>
+///   <argument_value> = <text>
```
wesleywiser, line 10: I wonder if it would be simpler to make the grammar:

```
<argument> = '\x1E' (<argument_name_value_pair> | <argument_value>)
<argument_name_value_pair> = <text> '\x1D' <text>
<argument_value> = <text>
```

Then we could always parse the leading `\x1E` and then do a string split on the `\x1D` if it exists.

wesleywiser: I think that would also make the format compatible with the current format. It's fine if we need to break that but we'll have to bump the format number and coordinate redeploying new versions on perf.rlo.

wesleywiser: It sounds like there might be another breaking change in the works from another contributor so if you think it's simpler to use the current strategy rather than this suggestion, that's fine with me!

nagisa:
I’m probably not going to pursue this further, but if anybody wants feel free to pick this up.

PR #123, opened 2020-08-09
In-memory construction of profiling data

Mark-Simulacrum:
perf.rust-lang.org is using this to avoid writing out profiles to disk when loading them over the network.


wesleywiser:
Thanks!

PR #124, opened 2020-08-16
Update README for changes in x.py defaults

jyn514:
See https://github.com/rust-lang/rust/pull/73964.

The choices are either to link to `stage1` with rustup or build stage 2; this recommends to build stage2 to be consistent with the existing docs.

### DIFF
--- a/summarize/Readme.md
+++ b/summarize/Readme.md
@@ -80,7 +80,7 @@ You can also profile your own custom build of rustc. First you'll have to clone 
 
 ```bash
 $ git clone https://github.com/rust-lang/rust.git
-$ ./x.py build
+$ ./x.py build --stage 2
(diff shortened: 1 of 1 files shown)


summarize/Readme.md
```diff
@@ -80,7 +80,7 @@ You can also profile your own custom build of rustc. First you'll have to clone
 
 ```bash
 $ git clone https://github.com/rust-lang/rust.git
-$ ./x.py build
+$ ./x.py build --stage 2
 # This will take a while...
 $ rustup toolchain link mytoolchain build/x86_64-unknown-linux-gnu/stage2
```
Mark-Simulacrum, line 85: Hm really stage1 should be enough; you should be comparing builds produced locally anyway.

PR #131, opened 2020-09-11
Remove the SerializationSink trait

michaelwoerister:
This PR removes the `SerializationSink` trait and thus the ability to plug in different serialization strategies at the client level. The `FileSerializationSink` is now the only implementation and is renamed to just `SerializationSink`.

This simplifies the public interface of the `measureme` crate.

r? @wesleywiser


measureme/src/serialization.rs
```diff
@@ -11,70 +14,172 @@ impl Addr {
...
-    where
-        W: FnOnce(&mut [u8]);
+#[derive(Debug)]
+pub struct SerializationSink {
+    data: Mutex<Inner>,
+}
+
+/// The `BackingStorage` is what the data gets written to.
+trait BackingStorage: Write + Send + Debug {
+    fn drain_bytes(&mut self) -> Vec<u8>;
```
wesleywiser, line 24: Should we #[cfg(test)] this or would that be too weird?

measureme/src/serialization.rs
```diff
@@ -11,70 +14,172 @@ impl Addr {
...
+            }),
         }
     }
 
+    /// Create a copy of all data written so far. This method meant to be used
+    /// for writing unit tests. It will panic if the underlying `BackingStorage`
+    /// does not implement `extract_bytes`.
     pub fn into_bytes(self) -> Vec<u8> {
-        self.data.into_inner()
+        let mut data = self.data.lock();
```
wesleywiser, line 65: I think this could be cfg'd then as well.

measureme/src/serialization.rs
michaelwoerister: I just tried it but it's complicated because `analyzeme` uses these things in its tests too -- but Cargo does not compile the dependencies of crate X with `--test` or `--cfg test` when it compiles crate X with `--test`. I suspect one can somehow use Cargo features to get this to work (although I couldn't find an example of this in the Cargo book).

wesleywiser: Ah ok. Let's add a doc comment to this saying it shouldn't be used except by tests because it will panic.

michaelwoerister: Done.

wesleywiser (approved):

wesleywiser:
Thanks!

PR #132, opened 2020-09-18
Implement paged data serialization

### LINKED ISSUE #128: Single file on disk format
Right now, `measureme` records data into three files for each profiling session: one for event data, one for string data and one which maps string ids to positions in the string data file. We've often gotten requests to consolidate this down to a single file. 

One idea I've been thinking about is to used fixed size pages to store the event data and string data:

```
--------------------------------------------------------------------
| event0 | event1 | eventN... |     stringN... | string1 | string0 |
--------------------------------------------------------------------
```

Events would be written from the beginning of the page to the end and strings would be written from the end of the page to the beginning. Once there is insufficient space in the page to write a new event or a new string, a new page would be created and the data would be written there. Of course, we'll want to know how many events there are and where the string data starts in the page, so we'll probably need a header block at the start of the page.

That takes care of two of the three data types but we still have the string index data. To handle that, we can have dedicated pages which just contain the string index data. 

The resulting data file might look a bit like this:

```
----page0-----------------------------------------------------------
| header | is_string_index=1 | event_count=0 | string_start_idx=0
| idx0 | idx1 | idxN... |
--------------------------------------------------------------------

----page1-----------------------------------------------------------
| header | is_string_index=0 | event_count=n | string_start_idx=x
| event0 | event1 | eventN... |     stringN... | string1 | string0 |
--------------------------------------------------------------------

----pageN-----------------------------------------------------------
...
--------------------------------------------------------------------
```

On top of this, there are quite a few enhancements we can add:

- String index pages can have a "forward pointer" to the next string index page so we don't have to scan every page linearly.
- Header blocks in the pages give us some scratch space to record additional data or metadata
- We can also change our approach to multi-threaded writing. For example, we can hand pages out per-thread which would then allow us to do concurrent writes without any locking since each thread is writing to different pages. Adding new pages could be done with an atomic add operation so no locks would be required anywhere. 

cc @michaelwoerister

michaelwoerister:
This PR implements the new paging-based data format, roughly as described in #128. However, the format has been tweaked to allow variable sized pages.

This is a work in progress:
- [x] Make `SerializationSink` paging-based
- [x] Write unit tests for `SerializationSink`
- [x] Add documentation about the data format
- [x] Add more explanatory comments throughout the `serialization` module
- [x] Add a proper file header to the monolithic file in addition to the three stream headers
- [x] Run local perf.rlo tests with a `rustc` using this version of `measureme`.

Closes #128


measureme/src/serialization.rs
```diff
@@ -279,6 +299,10 @@ impl SerializationSink {
 
         curr_addr
     }
+
+    pub fn as_std_write<'a>(&'a self) -> StdWriteAdapter<'a> {
```
wesleywiser, line 303:
```suggestion
    pub fn as_std_write<'a>(&'a self) -> impl Write {
```

analyzeme/src/profiling_data.rs
```diff
@@ -46,49 +45,60 @@ pub struct ProfilingData {
...
-                fs::read(paths.string_index_file).expect("couldn't read string_index file");
-            let event_data = fs::read(paths.events_file).expect("couldn't read events file");
-
-            ProfilingData::from_buffers(string_data, index_data, event_data)
+            let mut msg = format!(
+                "Could not find profiling data file `{}`.",
+                paged_path.display()
+            );
+
+            // Let's try to give a helpful error message if we encounter files
```
wesleywiser, line 69: 👍

wesleywiser (commented):
This is looking good to me!

wesleywiser:
This looks great to me! Do you have any feeling for the performance for this approach? Is it on par with the previous implementation?

michaelwoerister:
Yeah, I expect performance to be the same as the proof of concept implementation.

wesleywiser (approved):

PR #137, opened 2020-10-01
Allow StringTable to use the full 32-bit address space

michaelwoerister:
This PR changes the encoding for string references within the `StringTable` to use 5 bytes instead of 4 bytes. This way we can address 4 GB of data instead of 1 GB. The code becomes quite a bit simpler because we can remove a lot of manual bit-fiddling operations.

The only downside of the new approach is that string reference heavy data (e.g. from recording query keys) becomes about 5% larger. Between the larger address space support and the maintainability improvements this seems like an acceptable tradeoff to me. (I also have a half-done implementation of the same functionality without the additional space requirements -- but that turned out to make for such messy de-/serialization code that I abandoned it)

The size of profiling data that is collected without query keys should not be affected by this change.

r? @wesleywiser

### DIFF
--- a/measureme/src/stringtable.rs
+++ b/measureme/src/stringtable.rs
@@ -15,27 +15,24 @@
 //! The byte-level encoding of component lists uses the structure of UTF-8 in
 //! order to save space:
 //!
-//! - A valid UTF-8 codepoint never starts with the bits `10` as this bit
-//!   prefix is reserved for bytes in the middle of a UTF-8 codepoint byte
-//!   sequence. We make use of this fact by letting all string ID components
-//!   start with this `10` prefix. Thus when we parse the contents of a value
-//!   we know to stop if the start byte of the next codepoint has this prefix.
+//! - A valid UTF-8 codepoint never starts with the byte `0xFE`. We make use
+//!   of this fact by letting all string ID components start with this `0xFE`
+//!   prefix. Thus when we parse the contents of a value we know to stop if
+//!   we encounter this byte.
 //!
-//! - A valid UTF-8 string cannot contain the `0xFF` byte and since string IDs
-//!   start with `10` as described above, they also cannot start with a `0xFF`
-//!   byte. Thus we can safely use `0xFF` as our component list terminator.
+//! - A valid UTF-8 string cannot contain the `0xFF` byte. Thus we can safely
+//!   use `0xFF` as our component list terminator.
 //!
 //! The sample composite string ["abc", ID(42), "def", TERMINATOR] would thus be
 //! encoded as:
 //!
 //! ```ignore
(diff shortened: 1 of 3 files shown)


measureme/src/stringtable.rs
```diff
@@ -187,11 +180,10 @@ impl<'s> StringComponent<'s> {
                 &mut bytes[s.len()..]
             }
             StringComponent::Ref(string_id) => {
-                assert!(string_id.0 == string_id.0 & STRING_ID_MASK);
-                let tagged = string_id.0 | (1u32 << 31);
-
-                &mut bytes[0..4].copy_from_slice(&tagged.to_be_bytes());
-                &mut bytes[4..]
+                assert!(STRING_REF_ENCODED_SIZE == 5);
```
wesleywiser, line 183: Is this the assert we actually want here? (It's trivially true and rustc actually seems to optimize it out before codegen even occurs but I'm not sure why we'd assert that in this method.)

measureme/src/stringtable.rs
```diff
@@ -187,11 +180,10 @@ impl<'s> StringComponent<'s> {
...
+                bytes[0] = STRING_REF_TAG;
+                &mut bytes[1..5].copy_from_slice(&string_id.0.to_le_bytes());
+                &mut bytes[5..]
```
wesleywiser, line 186: Oh, I see it's for this. How do you feel about doing this instead? 

```suggestion
                &mut bytes[1..STRING_REF_ENCODED_SIZE].copy_from_slice(&string_id.0.to_le_bytes());
                &mut bytes[STRING_REF_ENCODED_SIZE..]
```

measureme/src/stringtable.rs
michaelwoerister: Yeah, the `assert!` is supposed to be a self-testing explanation of the magic numbers below. I think the best thing to do is (1) leaving the assertion in, (2) adding a comment that explains its purpose, and (3) applying the changes you suggest below.

For this kind of low-level byte shuffling stuff I like to add these kinds of tripwires that decrease the possibility of overlooking necessary adaptations when the encoding changes.

michaelwoerister: I added a comment but ended up *not* replacing the literal `5` with `STRING_REF_ENCODED_SIZE`. I don't know about you but writing `STRING_REF_ENCODED_SIZE` did not actually seem to make the code more readable.

What might make it more readable would be to have a bunch of constants like `STRING_REF_TAG_OFFSET`, `STRING_REF_VALUE_START_OFFSET`, and `STRING_REF_VALUE_END_OFFSET`, so that the encoding is kind of explained in one place. I don't think that's quite worth the trouble though.

wesleywiser: That seems reasonable to me 👍

wesleywiser (approved):

PR #138, opened 2020-10-06
Add a way to create an `EventId` with multiple arguments

wesleywiser:
cc @michaelwoerister

### DIFF
--- a/measureme/src/event_id.rs
+++ b/measureme/src/event_id.rs
@@ -78,4 +78,17 @@ impl<'p> EventIdBuilder<'p> {
             StringComponent::Ref(arg),
         ]))
     }
+
+    pub fn from_label_and_args(&self, label: StringId, args: &[StringId]) -> EventId {
+        // The capacity is the number of args + the same number for arg separators + the label.
+        let mut parts = Vec::with_capacity(args.len() * 2 + 1);
+        parts.push(StringComponent::Ref(label));
+
+        for arg in args {
+            parts.push(StringComponent::Value(SEPARATOR_BYTE));
+            parts.push(StringComponent::Ref(*arg));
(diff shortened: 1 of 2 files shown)


measureme/src/event_id.rs
```diff
@@ -78,4 +78,17 @@ impl<'p> EventIdBuilder<'p> {
             StringComponent::Ref(arg),
         ]))
     }
+
+    pub fn from_label_and_args(&self, label: StringId, args: &[StringId]) -> EventId {
+        // The capacity is the number of args + the same number for arg separators + the label.
+        let mut parts = Vec::with_capacity(args.len() * 2 + 1);
```
michaelwoerister, line 84: Looks like a good use case for a `SmallVec<[StringComponent; 8]>`. I'd be OK with adding a dependency on https://crates.io/crates/smallvec.

michaelwoerister:
Looks good!

Do we have any unit tests for this area?

measureme/src/event_id.rs
wesleywiser: Done! It was already in our dependency tree anyway so it shouldn't hurt our build times. 

I went ahead and used `7` instead of `8` since there's no situation where the `smallvec` will have 8 items in it.

wesleywiser:
There's a few tests here

https://github.com/rust-lang/measureme/blob/67920014f359d561294639e4b80c047461078b74/analyzeme/src/event.rs#L147-L177

measureme/src/event_id.rs
michaelwoerister: :+1:

michaelwoerister (approved):
Thanks, @wesleywiser!

PR #142, opened 2020-10-23
Decode paged buffers from memory

Mark-Simulacrum:
Do we expect to keep the compatibility for deserializing 0.7 (i.e., older) profile data for some time with newer analyzeme as well? That would be great from perf.rlo's side but we can depend on an older version too if needed.

r? @wesleywiser

### DIFF
--- a/analyzeme/src/profiling_data.rs
+++ b/analyzeme/src/profiling_data.rs
@@ -82,6 +82,18 @@ impl ProfilingData {
         }
     }
 
+    pub fn from_paged_buffer(data: Vec<u8>) -> Result<ProfilingData, Box<dyn Error + Send + Sync>> {
+        verify_file_header(&data, FILE_MAGIC_TOP_LEVEL, None, "top-level")?;
+
+        let mut split_data = measureme::split_streams(&data[FILE_HEADER_SIZE..]);
+
+        let string_data = split_data.remove(&PageTag::StringData).unwrap();
+        let index_data = split_data.remove(&PageTag::StringIndex).unwrap();
+        let event_data = split_data.remove(&PageTag::Events).unwrap();
+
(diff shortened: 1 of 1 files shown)


wesleywiser:
> Do we expect to keep the compatibility for deserializing 0.7 (i.e., older) profile data for some time with newer analyzeme as well? That would be great from perf.rlo's side but we can depend on an older version too if needed.

I don't think we have that compatibility currently since we've been bumping the `CURRENT_FILE_FORMAT_VERSION` number so I think `analyzeme` will panic if it tries to read old files.

Mark-Simulacrum:
Hm, but would the from_buffers function in this file always panic? Or just happens to work? We should probably delete it if it won't work...

wesleywiser:
Looks like it won't panic, but it will return an error since the file format versions don't match. Were you going to pipe `0.7` data into this function? 

https://github.com/rust-lang/measureme/blob/28b1eea82d7ed583f21193446b141a3052e816ec/measureme/src/file_header.rs#L66-L76

Mark-Simulacrum:
The function this PR adds is intended solely for the new file format. I was semi-hoping that given the 3 buffer taking function existing it would work for the older format too, which would obviously be less hassle on the other end -- but I'd be fine linking in two different analyzeme versions, that shouldn't be too hard.

wesleywiser:
Ah ok got it. No, I don't think that will work even if we bypassed the version check because we changed how strings get stored which fixed a failed assertion if you tried to profile really large crates like `rustc_middle`.

Mark-Simulacrum:
Thanks! I think this is still good to merge (we might want to adjust the 3 buffer taking function to be private since it seems likely useless now, though I'd prefer to do that in a separate PR).

wesleywiser (approved):

PR #143, opened 2020-11-03
Hardware performance counter support (via `rdpmc`).

eddyb:
*Note: this is a companion to https://github.com/rust-lang/rust/pull/78781, and duplicates some information with it for convenience*

## Credits

I'd like to start by thanking @alyssais, @cuviper, @edef1c, @glandium, @jix, @Mark-Simulacrum, @m-ou-se, @mystor, @nagisa, @puckipedia, and @yorickvP, for all of their help with testing, and valuable insight and suggestions.
Getting here wouldn't have been possible without you!

(If I've forgotten anyone please let me know, I'm going off memory here, plus some discussion logs)

## Summary

This PR adds support for counting hardware events such as "instructions retired" (as opposed to being limited to time measurements), using the `rdpmc` instruction on `x86_64` Linux.
Additionally, as `asm!` is used, `measureme` needs to built with nightly Rust, and with `features = ["nightly"]`, for the support to be enabled.

While other OSes may eventually be supported, preliminary research suggests some kind of kernel extension/driver is required to enable this, whereas on Linux any user can profile (at least) their own threads.

Supporting Linux on architectures other than x86_64 should be much easier (provided the hardware supports such performance counters), and was mostly not done due to a lack of readily available test hardware.
That said, 32-bit `x86` (aka `i686`) would be almost trivial to add and test once we land the initial `x86_64` version (as all the CPU detection code can be reused).

These new APIs were (backwards-compatibly) added to `measureme`:
* `counters` module holding `Counter` and all the counters implemented so far
    * attempting to create a counter may return `Err` if unsupported, but are not `#[cfg]`'d out at compile-time
* `Counter::by_name` creates a `Counter` based on its name
    * this is only to avoid needing to reimplement this in users of `measureme` which want to expose the counter names (e.g. `rustc` via a `-Z self-profile-counter` flag)
* `Profiler::with_counter` taking the `Counter` to use for event endpoints
    * the existing `Profiler::new` continues to default to measuring time (i.e. equivalent to `Profiler::with_counter(Counter::by_name("wall-time").unwrap())`)
    * a `Profiler` is currently limited to one `Counter` but long-term we could adjust the binary format to allow including several counters' values in a `measureme` event

Information about counters (name and units) is included in the JSON blob describing the profile data, which should still allow unupdated tools to read the new profiles (but they would report e.g. instructions as nanoseconds).
The backwards compatibility aspect was more important before the recent breaking change was introduced (for using just one file instead of 3, per profile), though avoiding breaking changes still seems good practice here.

The named counters so far are:
* `wall-time`: the existing time measurement
    * name chosen for consistency with `perf.rust-lang.org`
    * continues to use `std::time::Instant` for a nanosecond-precision "monotonic clock"
* `instructions:u`: the hardware performance counter usually referred to as "Instructions retired"
    * here "retired" (roughly) means "fully executed"
    * the `:u` suffix is from the Linux `perf` tool and indicates the counter only runs while userspace code is executing, and therefore counts no kernel instructions
        * *see [Caveats/Subtracting IRQs](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Subtracting-IRQs) for why this isn't entirely true and why `instructions-minus-irqs:u` should be preferred instead*
* `instructions-minus-irqs:u`: same as `instructions:u`, except the count of hardware interrupts ("IRQs" here for brevity) is subtracted
    * *see [Caveats/Subtracting IRQs](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Subtracting-IRQs) for why this should be preferred over `instructions:u`*
* `instructions-minus-r0420:u`: experimental counter, same as `instructions-minus-irqs:u` but subtracting an undocumented counter (`r0420:u`) instead of IRQs
    * the `rXXXX` notation is again from Linux `perf`, and indicates a "raw" counter, with a hex representation of the low-level counter configuration - this was picked because we still don't *really* know what it is
    * this only exists for (future) testing and isn't included/used in any comparisons/data we've put together so far
    * *see [Challenges/Zen's undocumented 420 counter](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Epilogue-Zen’s-undocumented-420-counter) for details on how this counter was found and what it does*

---

#### Write-up / report

Because of how extensive the full report ended up being, I've kept most of it [on `hackmd.io`](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view), but for convenient access, here are all the sections (with individual links):
<sup>(someone suggested I'd make a backup, so [here it is on the wayback machine](http://web.archive.org/web/20201103104237/https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view) - I'll need to remember to update that if I have to edit the write-up)</sup>

* [**Motivation**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Motivation)

* [**Results**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Results)
    * [**Overhead**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Overhead)
    *Preview (see the report itself for more details):*

    |Counter|Total<br>`instructions-minus-irqs:u`|Overhead from "Baseline"<br>(for all 1903881<br>counter reads)|Overhead from "Baseline"<br>(per each counter read)|
    |-|-|-|-|
    |Baseline|63637621286 ±6||
    |`instructions:u`|63658815885 ±2|&nbsp;&nbsp;+21194599 ±8|&nbsp;&nbsp;+11|
    |`instructions-minus-irqs:u`|63680307361 ±13|&nbsp;&nbsp;+42686075 ±19|&nbsp;&nbsp;+22|
    |`wall-time`|63951958376 ±10275|+314337090 ±10281|+165|

    * [**"Macro" noise (self time)**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#“Macro”-noise-(self-time))
    *Preview (see the report itself for more details):*

    || `wall-time` (ns) | `instructions:u` | `instructions-minus-irqs:u`
    -: | -: | -: | -:
    `typeck` | 5478261360 ±283933373 (±~5.2%) | 17350144522 ±6392 (±~0.00004%) | 17351035832.5 ±4.5 (±~0.00000003%)
    `expand_crate` | 2342096719 ±110465856 (±~4.7%) | 8263777916 ±2937 (±~0.00004%) | 8263708389 ±0 (±~0%)
    `mir_borrowck` | 2216149671 ±119458444 (±~5.4%) | 8340920100 ±2794 (±~0.00003%) | 8341613983.5 ±2.5 (±~0.00000003%)
    `mir_built` | 1269059734 ±91514604 (±~7.2%) | 4454959122 ±1618 (±~0.00004%) | 4455303811 ±1 (±~0.00000002%)
    `resolve_crate` | 942154987.5 ±53068423.5 (±~5.6%) | 3951197709 ±39 (±~0.000001%) | 3951196865 ±0 (±~0%)

    * [**"Micro" noise (individual sampling intervals)**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#“Micro”-noise-(individual-sampling-intervals))

* [**Caveats**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Caveats)
    * [**Disabling ASLR**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Disabling-ASLR)
    * [**Non-deterministic proc macros**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Non-deterministic-proc-macros)
    * [**Subtracting IRQs**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Subtracting-IRQs)
    * [**Lack of support for multiple threads**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Lack-of-support-for-multiple-threads)

* [**Challenges**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Challenges)
    * [**How do we even read hardware performance counters?**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#How-do-we-even-read-hardware-performance-counters)
    * [**ASLR: it's free entropy**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#ASLR-it’s-free-entropy)
    * [**The serializing instruction**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#The-serializing-instruction)
    * [**Getting constantly interrupted**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Getting-constantly-interrupted)
    * [**AMD patented time-travel and dubbed it `SpecLockMap`<br><sup>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;or: "how we accidentally unlocked `rr` on AMD Zen"</sup>**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#AMD-patented-time-travel-and-dubbed-it-SpecLockMapnbspnbspnbspnbspnbspnbspnbspnbspor-“how-we-accidentally-unlocked-rr-on-AMD-Zen”)
    * [**`jemalloc`: purging will commence in ten seconds**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#jemalloc-purging-will-commence-in-ten-seconds)
    * [**Rebasing *shouldn't* affect the results, right?**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Rebasing-*shouldn’t*-affect-the-results,-right)
    * [**Epilogue: Zen's undocumented 420 counter**](https://hackmd.io/sH315lO2RuicY-SEt7ynGA?view#Epilogue-Zen’s-undocumented-420-counter)

### DIFF
--- a/measureme/src/counters.rs
+++ b/measureme/src/counters.rs
@@ -11,8 +11,9 @@ macro_rules! really_warn {
 }
 
 pub enum Counter {
-    WallTime(WallTime),
-    Instructions(Instructions),
+    Zero(InstructionsMinusIrqs),
+    WallTime(InstructionsMinusIrqs, WallTime),
+    Instructions(InstructionsMinusIrqs, Instructions),
     InstructionsMinusIrqs(InstructionsMinusIrqs),
     InstructionsMinusRaw0420(InstructionsMinusRaw0420),
 }
@@ -20,12 +21,17 @@ pub enum Counter {
 impl Counter {
     pub fn by_name(name: &str) -> Result<Self, Box<dyn Error + Send + Sync>> {
         Ok(match name {
-            WallTime::NAME => Counter::WallTime(WallTime::new()),
-            Instructions::NAME => Counter::Instructions(Instructions::new()?),
-            InstructionsMinusIrqs::NAME => {
+            "0" => Counter::Zero(InstructionsMinusIrqs::new()?),
+            "t" | WallTime::NAME => {
+                Counter::WallTime(InstructionsMinusIrqs::new()?, WallTime::new())
+            }
+            "i" | Instructions::NAME => {
+                Counter::Instructions(InstructionsMinusIrqs::new()?, Instructions::new()?)
+            }
+            "I" | InstructionsMinusIrqs::NAME => {
                 Counter::InstructionsMinusIrqs(InstructionsMinusIrqs::new()?)
             }
-            InstructionsMinusRaw0420::NAME => {
+            "r" | InstructionsMinusRaw0420::NAME => {
                 Counter::InstructionsMinusRaw0420(InstructionsMinusRaw0420::new()?)
             }
             _ => return Err(format!("{:?} is not a valid counter name", name).into()),
@@ -34,11 +40,12 @@ impl Counter {
 
     pub(super) fn describe_as_json(&self) -> String {
         let (name, units) = match self {
-            Counter::WallTime(_) => (
+            Counter::Zero(_) => ("zero", "[]"),
+            Counter::WallTime(..) => (
                 WallTime::NAME,
                 r#"[["ns", 1], ["μs", 1000], ["ms", 1000000], ["s", 1000000000]]"#,
             ),
-            Counter::Instructions(_) => (Instructions::NAME, r#"[["instructions", 1]]"#),
+            Counter::Instructions(..) => (Instructions::NAME, r#"[["instructions", 1]]"#),
             Counter::InstructionsMinusIrqs(_) => {
                 (InstructionsMinusIrqs::NAME, r#"[["instructions", 1]]"#)
             }
@@ -52,8 +59,9 @@ impl Counter {
     #[inline]
     pub(super) fn since_start(&self) -> u64 {
         match self {
-            Counter::WallTime(counter) => counter.since_start(),
-            Counter::Instructions(counter) => counter.since_start(),
+            Counter::Zero(_) => 0,
+            Counter::WallTime(_, counter) => counter.since_start(),
+            Counter::Instructions(_, counter) => counter.since_start(),
             Counter::InstructionsMinusIrqs(counter) => counter.since_start(),
             Counter::InstructionsMinusRaw0420(counter) => counter.since_start(),
         }
@@ -129,6 +137,17 @@ impl InstructionsMinusIrqs {
     fn since_start(&self) -> u64 {
         let (instructions, irqs) = (&self.instructions, &self.irqs).read();
         instructions.wrapping_sub(irqs).wrapping_sub(self.start)
+    }
+}
+
+// HACK(eddyb) dump total `instructions-minus-irqs:u` for statistics.
+impl Drop for InstructionsMinusIrqs {
+    fn drop(&mut self) {
+        eprintln!(
+            "pid={:06} instructions-minus-irqs:u={}",
+            std::process::id(),
+            self.since_start(),
+        );
     }
 }
 
--- a/measureme/src/counters.rs
+++ b/measureme/src/counters.rs
@@ -81,7 +81,7 @@
 //!     which can cause non-deterministic overcounting for instructions following
 //!     an atomic instruction (such as found in heap allocators, or `measureme`)
 //!     * this is automatically detected, with a `log` message pointing the user
-//!       to [https://github.com/mozilla/rr/wiki/Zen] for guidance on how to
+//!       to <https://github.com/mozilla/rr/wiki/Zen> for guidance on how to
 //!       disable `SpecLockMap` on their system (sadly requires root access)
 //!
 //! Even if some of the above caveats apply for some profiling setup, as long as
--- a/measureme/src/counters.rs
+++ b/measureme/src/counters.rs
@@ -11,9 +11,8 @@ macro_rules! really_warn {
 }
 
 pub enum Counter {
-    Zero(InstructionsMinusIrqs),
-    WallTime(InstructionsMinusIrqs, WallTime),
-    Instructions(InstructionsMinusIrqs, Instructions),
+    WallTime(WallTime),
+    Instructions(Instructions),
     InstructionsMinusIrqs(InstructionsMinusIrqs),
     InstructionsMinusRaw0420(InstructionsMinusRaw0420),
 }
@@ -21,17 +20,12 @@ pub enum Counter {
 impl Counter {
     pub fn by_name(name: &str) -> Result<Self, Box<dyn Error + Send + Sync>> {
         Ok(match name {
-            "0" => Counter::Zero(InstructionsMinusIrqs::new()?),
-            "t" | WallTime::NAME => {
-                Counter::WallTime(InstructionsMinusIrqs::new()?, WallTime::new())
-            }
-            "i" | Instructions::NAME => {
-                Counter::Instructions(InstructionsMinusIrqs::new()?, Instructions::new()?)
-            }
-            "I" | InstructionsMinusIrqs::NAME => {
+            WallTime::NAME => Counter::WallTime(WallTime::new()),
+            Instructions::NAME => Counter::Instructions(Instructions::new()?),
+            InstructionsMinusIrqs::NAME => {
                 Counter::InstructionsMinusIrqs(InstructionsMinusIrqs::new()?)
             }
-            "r" | InstructionsMinusRaw0420::NAME => {
+            InstructionsMinusRaw0420::NAME => {
                 Counter::InstructionsMinusRaw0420(InstructionsMinusRaw0420::new()?)
             }
             _ => return Err(format!("{:?} is not a valid counter name", name).into()),
@@ -40,12 +34,11 @@ impl Counter {
 
     pub(super) fn describe_as_json(&self) -> String {
         let (name, units) = match self {
-            Counter::Zero(_) => ("zero", "[]"),
-            Counter::WallTime(..) => (
+            Counter::WallTime(_) => (
                 WallTime::NAME,
                 r#"[["ns", 1], ["μs", 1000], ["ms", 1000000], ["s", 1000000000]]"#,
             ),
-            Counter::Instructions(..) => (Instructions::NAME, r#"[["instructions", 1]]"#),
+            Counter::Instructions(_) => (Instructions::NAME, r#"[["instructions", 1]]"#),
             Counter::InstructionsMinusIrqs(_) => {
                 (InstructionsMinusIrqs::NAME, r#"[["instructions", 1]]"#)
             }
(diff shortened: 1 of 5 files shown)


measureme/src/counters.rs
```diff
@@ -0,0 +1,869 @@
...
+}
+
+pub struct InstructionsMinusIrqs {
+    instructions: hw::Counter,
+    irqs: hw::Counter,
+    start: u64,
+}
+
+impl InstructionsMinusIrqs {
+    const NAME: &'static str = "instructions-minus-irqs:u";
```
bjorn3, line 113: Maybe use `instructions:u` as name here for the real amount of userspace instructions and use something like `instructions:u+int` as name for the amount of userspace instructions combined with any bogus counted instructions caused by interrupt returns or similar things.

eddyb: If we do that I'd be against using `instructions:u` as the name for anything, because `perf stat` *does not* do any subtracting, and we used the `instructions:u` name specifically for `perf stat` familiarity (which is why `perf.rust-lang.org` also uses it).

Furthermore, we might want to try and get `perf.rust-lang.org` to record the right counter and *also* show `instructions-minus-irqs:u`, so ideally we'd use different names to tell them apart.

But I agree the current setup is not the most user-friendly, it was just the easiest way to keep things clear while we worked on this.

AlphaHot (approved):

ambiso:
Hi,

I'm not sure if this is directly relevant to this PR,
and I very well may not be telling anyone anything new.

Still, since I saw the insistence to remove all noise from the measurements I just wanted to make sure, that people are aware of research that suggests many optimizations performed by LLVM on `-O3` are for naught when tested with random stack and binary layout [1,2,3].
This would suggest that you would actually not want to simply disable ASLR, but test on every possible layout.

If this is the wrong pull request to post this comment, or you are aware of this research, or that it doesn't matter for what you're trying to achieve, please feel free to ignore this comment, move it, or whatever.

Take care,
ambiso

[1]: https://users.cs.northwestern.edu/~robby/courses/322-2013-spring/mytkowicz-wrong-data.pdf
[2]: https://youtu.be/r-TLSBdHe1A?t=948
[3]: https://github.com/ccurtsinger/stabilizer

eddyb:
@ambiso The way ASLR feeds into this is roughly `HashMap<*const T, ...>` (the equivalent of `unordered_map<T*, ...>` in C++), i.e. hashing pointer addresses, meaning the choice of addresses affects the order in the hash map, and therefore the performance characteristics, by somewhere around `±0.01%`.

I'm not sure how optimizations, or *stack layout* feeds into it, but we're not (yet) measuring anything like cache effectiveness, the entire problem with "ASLR non-determinism" in our case is our fault for relying on (interned) pointer addresses for cheaper hashing (and avoiding having to *store* a hash somewhere at all).

measureme/src/counters.rs
```diff
@@ -0,0 +1,869 @@
...
+
+            if index == 0 {
+                return Err(format!(
+                    "perf_event_mmap_page: no allocated hardware register (ran out?)"
+                )
+                .into());
+            }
+            counter.reg_idx = index - 1;
+
+            if (cfg!(not(accurate_seqlock_rdpmc)) || true) && pmc_width != 48 {
```
wesleywiser, line 303: I think this can just be 

```suggestion
            if (cfg!(not(accurate_seqlock_rdpmc))) && pmc_width != 48 {
```

measureme/src/counters.rs
```diff
@@ -0,0 +1,869 @@
...
+        /// has updated one or more fields within the mmap page).
+        #[inline]
+        fn access_mmap_page_with_seqlock<T>(
+            &self,
+            attempt: impl Fn(&perf_event_mmap_page) -> T,
+        ) -> T {
+            // FIXME(eddyb) it's probably UB to use regular reads, especially
+            // from behind `&T`, with the only synchronization being barriers.
+            // Probably needs atomic reads, and stronger ones at that, for the
+            // `lock` field, than the fields (which would be `Relaxed`?).
```
wesleywiser, line 325: Do you think this should be fixed before merging?

measureme/src/raw_event.rs
```diff
@@ -12,40 +12,44 @@ pub struct RawEvent {
...
 }
 
-/// `RawEvents` that have an end time stamp with this value are instant events.
-const INSTANT_TIMESTAMP_MARKER: u64 = 0xFFFF_FFFF_FFFF;
+/// `RawEvents` that have an end counter value with this value are instant events.
+const INSTANT_COUNT_MARKER: u64 = 0xFFFF_FFFF_FFFF;
 
-/// The max instant timestamp we can represent with the 48 bits available.
+/// The max instant counter value we can represent with the 48 bits available.
+// FIXME(eddyb) s/TIMESTAMP/COUNT/
```
wesleywiser, line 28: This is a note for me to create issues to clean this up at the next major version.

measureme/src/counters.rs
```diff
@@ -0,0 +1,869 @@
...
+                // FIXME(eddyb) the Intel and AMD manuals warn about the need for
+                // "serializing instructions" before/after `rdpmc`, if avoiding any
+                // reordering is desired, but do not agree on the full set of usable
+                // "serializing instructions" (e.g. `mfence` isn't listed in both).
+                //
+                // The only usable, and guaranteed to work, "serializing instruction"
+                // appears to be `cpuid`, but it doesn't seem easy to use, especially
+                // due to the overlap in registers with `rdpmc` itself, and it might
+                // have too high of a cost, compared to serialization benefits (if any).
+                asm!("rdpmc", in("ecx") reg_idx, out("eax") lo, out("edx") hi, options(nostack));
```
wesleywiser, line 417: Do we need this dead code? I think we can probably just move this comment down to the else block.

measureme/src/counters.rs
```diff
@@ -0,0 +1,869 @@
...
+    }
+
+    impl super::HwCounterRead for (&Counter, &Counter) {
+        type Output = (u64, u64);
+
+        #[inline]
+        fn read(&self) -> (u64, u64) {
+            // HACK(eddyb) keep the accurate code around while not using it,
+            // to minimize overhead without losing the more complex implementation.
+            if (cfg!(accurate_seqlock_rdpmc) || cfg!(unserialized_rdpmc)) && false {
```
wesleywiser, line 382: I think we should document these cfg flags somewhere, perhaps in the module docs.

measureme/src/counters.rs
eddyb: The `|| true` are so that the `cfg`s are actually ignored, since I didn't want them to accidentally become `measureme` toggles.

measureme/src/counters.rs
eddyb: It's not used during actual counter reads, only once at the start, and it's the same as the C code examples, so this is more "theoretical", I think?

measureme/src/counters.rs
eddyb: The `|| true` (or `&& false`, here) are so that they're not actually exposed, see also https://github.com/rust-lang/measureme/pull/143#discussion_r522040961

measureme/src/counters.rs
eddyb: I found it useful to have several different versions around, for quick testing (without having to dig up `git reflog` or w/e), but I'm not sure how to best control it. Maybe a trait with associated `const`s - even if it's not publicly exposed for now.

wesleywiser (approved):

PR #146, opened 2020-11-20
Cleanup intra-doc links

jyn514:
- Remove unnecessary explicit links
- Remove `deny(intra_doc_resolution_failure)`: it's both renamed and on by default, so it will be caught by `deny(warnings)`.


wesleywiser (approved):
This is awesome! Thanks 👍

PR #147, opened 2020-11-23
Minor tweaks to the measureme::counters module.

eddyb:
I meant to get these done before landing #143, but that happened too soon. Will keep this as draft until ready.

The inline `asm!` changes I've validated by comparing the old and new versions (they produce identical assembly in godbolt set to "Intel dialect", i.e. the AT&T dialect gets parsed and rewritten to the Intel dialect I originally used).


wesleywiser (approved):
This looks great to me. Please ping me when you're ready for it to be merged.

eddyb:
Oops forgot about this but I think it was landable since https://github.com/rust-lang/rust/pull/78781#issuecomment-731944971.

wesleywiser (approved):

PR #148, opened 2020-12-06
add to invocation count for blocked events

andjo403:
as the time for blocked events is increased think that also the invocation count shall be increased.

was confused when looking at the result from https://github.com/rust-lang/rust/pull/79706 where parallell compiler is tested,
some queries was taking long time when running in parallell so was looking in the query after the problem as the invocation count was 1, but in many cases the problem was that the same query was executed in multiple threads at the same time resulting in blocked query events.


wesleywiser (approved):
Good catch!

PR #155, opened 2021-03-04
Crox: Collect mm_profiledata files in directory

### LINKED ISSUE #154: crox --dir generates empty json array
I might be doing something incorrectly, but maybe not. I've been following the [original article over at brlo](https://blog.rust-lang.org/inside-rust/2020/02/25/intro-rustc-self-profile.html), which generated 371 `mm_profiledata` files (a total of around 4GB of data, including event args) into my `profiles` folder. However, running `crox --dir profiles` from the project root (i.e. pointing at the correct directory), the resulting `chrome_profiler.json` contains only `[]`.

Converting a single profiledata file works as expected.

I'm using crox `9.1.0` with rustc -vV:
```
rustc 1.52.0-nightly (476acbf1e 2021-03-03)
binary: rustc
commit-hash: 476acbf1e9965b5e95c90f0d7d658709812b7003
commit-date: 2021-03-03
host: x86_64-unknown-linux-gnu
release: 1.52.0-nightly
LLVM version: 11.0.1
```

bugadani:
Hopefully closes #154 

Untested, but I think should be fine. Previously, `crox --dir <dir>` was looking for `.events` files, while `ProfileData` (and AFAICS everything else) works with `mm_profiledata`. This PR updates the file extension filter so that `crox --dir` may generate results without having to rename the world.

### DIFF
--- a/crox/src/main.rs
+++ b/crox/src/main.rs
@@ -5,6 +5,7 @@ use std::path::PathBuf;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use analyzeme::{ProfilingData, Timestamp};
+use measureme::file_header::FILE_EXTENSION;
 
 use serde::ser::SerializeSeq;
 use serde::{Serialize, Serializer};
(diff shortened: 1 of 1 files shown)


wesleywiser (approved):
Thank you!

PR #166, opened 2021-06-25
Handle new self-profiling information for incr-comp hash fingerprints

rylev:
Relies on changes in https://github.com/rust-lang/rust/pull/86619 where you can read a bit more about the goals of this change.

Putting into draft mode until https://github.com/rust-lang/rust/pull/86619 is more fully discussed.

### DIFF
--- a/summarize/src/analysis.rs
+++ b/summarize/src/analysis.rs
@@ -171,10 +171,15 @@ pub fn perform_analysis(data: ProfilingData) -> Results {
                     record_event_data(
                         &current_top.label,
                         &|data| match &current_top.event_kind[..] {
-                            QUERY_EVENT_KIND
-                            | GENERIC_ACTIVITY_EVENT_KIND
-                            | INCREMENTAL_HASH_RESULT_EVENT_KIND => {
+                            QUERY_EVENT_KIND | GENERIC_ACTIVITY_EVENT_KIND => {
                                 data.self_time -= current_event_duration;
+                            }
+                            INCREMENTAL_RESULT_HASHING_EVENT_KIND => {
+                                // We are within hashing something. If we now encounter something
+                                // within that event (like the nested "intern-the-dep-node" event)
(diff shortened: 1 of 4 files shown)


summarize/src/analysis.rs
```diff
@@ -171,7 +171,9 @@ pub fn perform_analysis(data: ProfilingData) -> Results {
                     record_event_data(
                         &current_top.label,
                         &|data| match &current_top.event_kind[..] {
-                            QUERY_EVENT_KIND | GENERIC_ACTIVITY_EVENT_KIND => {
+                            QUERY_EVENT_KIND
+                            | GENERIC_ACTIVITY_EVENT_KIND
+                            | INCREMENTAL_HASH_RESULT_EVENT_KIND => {
```
michaelwoerister, line 176: I took another look and I think we need to handle this the same way that we handle `INCREMENTAL_LOAD_RESULT_EVENT_KIND` below, just with `data.incremental_hashing_time -= current_event_duration;`. Why? Because in this step we take care of adjusting the timings of the thing currently on the stack. If a `INCREMENTAL_HASH_RESULT_EVENT_KIND` is on the stack it means we are within hashing something. If we now encounter something within that event (like the nested "intern-the-dep-node" event) then we don't want to attribute that to the hashing time.

michaelwoerister (commented):
I left a comment on how to timing adjustments for the new event.

rylev:
This should be ready, but leaving it as a draft since it relies on the changes to rustc landing first.

michaelwoerister:
I think this is ready to be merged, right? There is no harm in being able to handle the new events before those events actually get generated. Correct me if I'm wrong, @rylev `:)` 

r? @wesleywiser

rylev:
This is ready to review. I marked it as work-in-progress, because I wasn’t sure if the support in rustc needed to land first. Either way that support is being merged, so I’ll mark this as ready to review.

wesleywiser (approved):

PR #170, opened 2021-08-16
Fix wrong unreachable code in summarize aggregate

rylev:
This incorrect code was snuck in in #169. The code as it stands now should be exactly equivalent to how `summarize aggregate`  
worked before #169 landed. We are filtering out the newly "integer" type events and so the code can assume (as it did before) that it's only dealing with time related events.

r? @michaelwoerister


michaelwoerister (approved):
Thanks, @rylev!

PR #175, opened 2021-09-06
Check the artifact size event kind in summarize

rylev:
This makes it so that in summarize we only add to the artifact size if that event label is "ArtifactSize"

### DIFF
--- a/summarize/src/analysis.rs
+++ b/summarize/src/analysis.rs
@@ -254,7 +254,9 @@ pub fn perform_analysis(data: ProfilingData) -> Results {
                 thread.stack.push(current_event)
             }
             EventPayload::Integer(value) => {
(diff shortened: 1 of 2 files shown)


summarize/src/analysis.rs
```diff
@@ -254,7 +254,9 @@ pub fn perform_analysis(data: ProfilingData) -> Results {
                 thread.stack.push(current_event)
             }
             EventPayload::Integer(value) => {
-                artifact_sizes.push(ArtifactSize::new(current_event.label.into_owned(), value))
+                if current_event.event_kind == measureme::rustc::ARTIFACT_SIZE_EVENT_KIND {
```
michaelwoerister, line 257: We should be able to use `ARTIFACT_SIZE_EVENT_KIND` without the qualification since we already import `measureme::rustc::*` at the top of this file.

michaelwoerister (approved):

Issue #178, opened 2021-09-07
Make it easier to transition between versions


rylev:
Currently the process of moving from one version of measureme, summarize, and others to a newer version can be error prone. Generally newer versions of the tools emit and read different versions of the on-disk `.mm_profdata` file format. 

When reading a the profile data file, the parser first checks that the version of the file is exactly equal to the version the tool expects. https://github.com/rust-lang/measureme/blob/b9cccd7ad4c859a5e0e3dd6bff3daac7a190bdd7/measureme/src/file_header.rs#L77

If it's not equal the parser returns an error. 

In the end, we ideally want so that users rarely encounter errors, and when they do they instructions for fixing the error are simple and reliable. 

There are several improvements that we can do to address this issue.

## Better error messages

The most lightweight improvement would be to improve error messages when the tool encounters a file format version that it does not expect. 

If we detect the file version is less than the current version, tell users that they should regenerate the file if possible or install an older version of the tool that supports that file version. We could optionally, keep track of which versions of measure and co. work with which versions of the file format, and report to the user which one they should use. 

If we detect the file version is greater than the latest version, we can encourage users to upgrade the tools to the latest version. 

## Make the parser understand older file versions

Requiring the parser to understand at least the previous version of the file format will make the scenario where the user upgrades or installs the newest version of the measureme tool suite, but rustc has not yet been updated to emit the new file version. 

When changes to the file format are relatively small, this should be fairly trivial to do, but it may prove difficult if we ever need to dramatically change the file format. 

This does not help in the situation where the tool is older than the file format which can happen when users go longer periods between using the measureme tool suite. 

## Other considerations

* [**meta-executable**](https://github.com/rust-lang/measureme/issues/164): depending on how implement a meta-executable for the measureme tool suite, we could consider making it version aware and have it "fallback" to older versions when parsing fails. 
* perf.rlo [currently tries](https://github.com/rust-lang/rustc-perf/blob/d7f85aa72b2c648233e8513f1af85e697e32e726/collector/src/rustc-fake.rs#L345-L367) to use the `summarize` binary on its path (which is 9.1.2 currently) and when it fails it falls back version 9.0. If that then fails, everything fails. This hack works for now, but will require updating as future releases of measureme come out. Ideally perf.rlo shouldn't need to worry about this.

Mark-Simulacrum:
[...]

PR #179, opened 2021-09-10
Backport decodeme crate to v9.1.x

michaelwoerister:
This PR adds the `decodeme` crate to 9.1.x so that we can use it to make 10.x+ releases backwards compatible with the data format used in 9.x versions.

See https://github.com/rust-lang/measureme/issues/178 for more information.

r? @rylev & @wesleywiser


michaelwoerister:
Hm, I think technically the changes in here are not backwards compatible with `9.1.2` because `LightweightEvent` has changed. So maybe the crate version should not be `9.1.2` but something else? Maybe something unstable such as `0.9.0-decodeme-backport`?

analyzeme/src/profiling_data.rs
```diff
@@ -261,6 +169,11 @@ impl ProfilingDataBuilder {
         )
         .unwrap();
 
+        string_table.alloc_metadata(&*format!(
```
rylev, line 172: What's this for?

analyzeme/src/profiling_data.rs
```diff
@@ -82,97 +44,43 @@ impl ProfilingData {
...
+        self.event_decoder.decode_full_event(event_index)
+    }
 
-        LightweightEvent {
-            data: self,
-            event_index,
-            timestamp,
-            thread_id: raw_event.thread_id,
-        }
+    fn decode_lightweight_event<'a>(&'a self, event_index: usize) -> LightweightEvent {
```
rylev, line 82: nit: remove the lifetime

rylev (approved):
Looks good! I'm not sure how I feel about the backport strategy. Effectively we're making a new release of of v9.1 so that 10.0 can use that as a snapshot even though we don't intend for users to use the new version of 9.1. This feels strange.

Technically to be semver compliant we'd need to release this snapshot as 10.0 and then release 11.0 (since both introduce breaking changes).

analyzeme/src/profiling_data.rs
michaelwoerister: The string table contains a "metadata" entry that stores some global information. The `start_time` in there is used for mapping the timings in each event to a `SystemTime`, so we need it for test cases that look at those times.

michaelwoerister:
Closing in favor of https://github.com/rust-lang/measureme/pull/180.

PR #180, opened 2021-09-13
Prepare 9.2.0 release

michaelwoerister:
Makes a couple of methods in `ProfilingData` public in order to make it easier to start supporting older file formats in v10.0.0.

r? @rylev & @wesleywiser

### DIFF
--- a/analyzeme/src/profiling_data.rs
+++ b/analyzeme/src/profiling_data.rs
@@ -132,7 +132,7 @@ impl ProfilingData {
         event_byte_count / RAW_EVENT_SIZE
     }
 
-    pub(crate) fn decode_full_event<'a>(&'a self, event_index: usize) -> Event<'a> {
+    pub fn decode_full_event<'a>(&'a self, event_index: usize) -> Event<'a> {
         let event_start_addr = event_index_to_addr(event_index);
         let event_end_addr = event_start_addr.checked_add(RAW_EVENT_SIZE).unwrap();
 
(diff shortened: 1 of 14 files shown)


rylev (approved):
Looks great! Technically I would say we should update the README to point the install to this new version, but if we're going to release v10.0.0 right after this, it's not a blocker.

analyzeme/src/profiling_data.rs
```diff
@@ -174,6 +174,19 @@ impl ProfilingData {
             thread_id: raw_event.thread_id,
         }
     }
+
+    // This method is added only in order to make it easier to fix
+    // https://github.com/rust-lang/measureme/issues/173 in the future.
+    //
+    // It is marked as "internal" because we want to avoid that downstream tooling
+    // depends on RawEvent directly. Do not use this method directly.
```
rylev, line 182: Hmm this makes me a bit worried. If we don't strictly needed it right perhaps we should just not include it in this release?

wesleywiser (approved):
We should also update the `CHANGELOG.md` for this release.

analyzeme/src/profiling_data.rs
michaelwoerister: I don't have a strong opinion about this. I can remove it if you prefer.

michaelwoerister:
> We should also update the CHANGELOG.md for this release.

Done.

analyzeme/src/profiling_data.rs
michaelwoerister: Removed

PR #183, opened 2021-09-16
Test that ProfileData can read data from multiple file versions

rylev:
This includes real-world profiling data. The files are quite large so we might want to consider trimming them down before merging. 

r? @michaelwoerister


bjorn3:
Yes, please trim them to at most a couple kilobytes before merging.

rylev:
Got it to 348kb - I'm guessing we'll want more, but that will require editing page sizes which is a bit more involved.

bjorn3:
If you only have a couple events in the files would the files mostly consist of zeros? A lot of zeros in a row is easily compressable down to just a couple of bytes. You could let git compress them when creating packfiles or more reliably directly store them compressed.

rylev:
The data is real so it's thousands of real events. I could write a program to generate a file, and maybe that's what I'll do, but my original intent was to start with real-world data.

michaelwoerister:
Could we use something like https://crates.io/crates/flate2 to decompress the data in memory before using it?

michaelwoerister (commented):
Hm, the tests seem a bit opaque to me and decoding just the first event does not exercise the decoder a lot. 

I think either we want 
- a description of what is in the test file and then make sure we find certain events in some expected order, or
- we re-use and extend the code in `analyzeme::testing_common` to generate test data for a given format on-the-fly -- although that would mean that we need to do another v9.x release for making some functions in there public and move `testing_common` to decodeme for v10.

analyzeme/src/profiling_data.rs
```diff
@@ -434,6 +437,76 @@ mod tests {
...
+
+        assert_eq!(grouped_events["GenericActivity"].len(), 6429);
+        assert_eq!(grouped_events["IncrementalResultHashing"].len(), 2235);
+        assert_eq!(grouped_events["Query"].len(), 2258);
+
+        assert_eq!(&*grouped_events["GenericActivity"][6429 / 2].label, "metadata_decode_entry_implementations_of_trait");
+        assert_eq!(grouped_events["GenericActivity"][6429 / 2].duration(), Some(Duration::from_nanos(1200)));
+
+        assert_eq!(&*grouped_events["IncrementalResultHashing"][2235 - 1].label, "item_children");
+        assert_eq!(grouped_events["IncrementalResultHashing"][2235 - 1].duration(), Some(Duration::from_nanos(213003100)));
```
michaelwoerister, line 477: `Duration::from_nanos(213003100)` -- is that number correct? 😮

rylev: Looks like it. From mmview:

```
{
    kind: IncrementalResultHashing,
    label: item_children,
    additional_data: [],
    payload: 26415 μs - 239418 μs,
    thread_id: 2,
}
```

Is there some large incremental hashing job that happens near the end of the compilation?

michaelwoerister (approved):
Thanks a lot, @rylev!

PR #185, opened 2021-10-03
Avoid using ebx as an asm! operand

andjo403:
when I tried to rebase https://github.com/rust-lang/rust/pull/78781 I noticed that due to https://github.com/rust-lang/rust/pull/84658 this code did not compile any longer.

also due to https://github.com/rust-lang/rust/pull/83387 is merged we can use the intel asm and remove the duplication of all asm instructions

### DIFF
--- a/measureme/src/counters.rs
+++ b/measureme/src/counters.rs
@@ -525,45 +525,22 @@ mod hw {
     /// the width of the register (32 to 64 bits, e.g. 48-bit seems common).
     #[inline(always)]
     fn rdpmc(reg_idx: u32) -> u64 {
-        let (lo, hi): (u32, u32);
-        unsafe {
-            // NOTE(eddyb) below comment is outdated (the other branch uses `cpuid`).
-            if cfg!(unserialized_rdpmc) && false {
-                // FIXME(eddyb) the Intel and AMD manuals warn about the need for
-                // "serializing instructions" before/after `rdpmc`, if avoiding any
-                // reordering is desired, but do not agree on the full set of usable
-                // "serializing instructions" (e.g. `mfence` isn't listed in both).
-                //
-                // The only usable, and guaranteed to work, "serializing instruction"
-                // appears to be `cpuid`, but it doesn't seem easy to use, especially
-                // due to the overlap in registers with `rdpmc` itself, and it might
-                // have too high of a cost, compared to serialization benefits (if any).
-                asm!("rdpmc", in("ecx") reg_idx, out("eax") lo, out("edx") hi, options(nostack));
-            } else {
-                asm!(
-                    // Dummy `cpuid(0)` to serialize instruction execution.
-                    "xor eax, eax",
-                    // LLVM sometimes reserves `ebx` for its internal use, we so we need to use
-                    // a scratch register for it instead.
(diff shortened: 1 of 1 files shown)


measureme/src/counters.rs
```diff
@@ -543,16 +543,20 @@ mod hw {
                 asm!(
                     // Dummy `cpuid(0)` to serialize instruction execution.
                     "xor eax, eax",
+                    // LLVM sometimes reserves `ebx` for its internal use, we so we need to use
+                    // a scratch register for it instead.
+                    "mov {tmp_rbx:r}, rbx",
                     "cpuid",
+                    "mov rbx, {tmp_rbx:r}",
```
eddyb: I've noticed this. This is frustrating, as it makes `cpuid` undesirably troublesome (I wish Intel and AMD would agree on e.g. `mfence`/`lfence` being serializing). I suspect that using a LLVM intrinsic for `cpuid` would result in shorter codegen, can you try this? Either way, can you PM me on Zulip?

I've let that `rustc` PR sit, sadly (long story short, other factors forced us to abandon the grant it was part of), and it would be good to take stock of what's useful and what's just hacks we only needed to get the results we wanted.

measureme/src/counters.rs
```diff
@@ -572,42 +549,49 @@ mod hw {
...
+    #[inline(always)]
+    fn unserialized_rdpmc(reg_idx: u32) -> u64 {
+        let (lo, hi): (u32, u32);
+        unsafe {
+            asm!(
+                "rdpmc",
+                in("ecx") reg_idx,
+                lateout("eax") lo,
+                lateout("edx") hi,
+                options(nostack)
```
eddyb, line 591: I suppose this is good because it gives LLVM just one instruction and all of the register-wiring insight it needs.

measureme/src/counters.rs
```diff
@@ -572,42 +549,49 @@ mod hw {
...
     fn rdpmc_pair(a_reg_idx: u32, b_reg_idx: u32) -> (u64, u64) {
-        let (a_lo, a_hi): (u32, u32);
-        let (b_lo, b_hi): (u32, u32);
+        serialize_instruction_execution();
+        (unserialized_rdpmc(a_reg_idx), unserialized_rdpmc(b_reg_idx))
+    }
+
+    /// Dummy `cpuid(0)` to serialize instruction execution.
+    #[inline(always)]
+    fn serialize_instruction_execution() {
```
eddyb, line 558: Should be able to take this `fn` in isolation and try it on godbolt, for the comparison with `std::arch::x86_64::__cpuid(0);`.

eddyb (commented):
Had to review one commit at a time, but this makes sense. Sorry I haven't gone back to this yet. I ran into the `ebx` thing while trying to get the standalone "`SpecLockMap` checker" compiling on recent nightlies, for someone, and this looks like the correct fix.

There's only one concern I have (about the `cpuid(0)` code size with the `ebx` workaround), for which I left comments.

measureme/src/counters.rs
andjo403: minor difference with using arch version one extra xor and xchg instead of the mov that I used.

https://godbolt.org/z/8jv41bYsq

PR #189, opened 2021-10-07
Change install instructions to use stable branch

### LINKED ISSUE #182: Make installation instructions stable
https://github.com/rust-lang/measureme/pull/177 changed installation instructions to a specific tag which is much better than making them install whatever is on the main branch at the moment. However, it would be even better if the installation string could stay immutable regardless of what the current version is.

Now that we support old file formats too, we could introduce a "stable" branch that always points to the latest release. That would enable us to make the installation string something like:

```bash
cargo install --git https://github.com/rust-lang/measureme --branch stable crox
``` 

We could then also update the instructions in https://blog.rust-lang.org/inside-rust/2020/02/25/intro-rustc-self-profile.html, which is still a common entry point to self-profiling, I think (at least I always point folks to that).

rylev:
We now have a [stable](https://github.com/rust-lang/measureme/tree/stable) branch which points to the v10.0.0. We can now encourage folks to install off this branch instead of from a specific version or from master. 

(I also updated the READMEs with a few small fixes) 

r? @michaelwoerister 

Fixes #182

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# measureme [![Build Status](https://travis-ci.com/rust-lang/measureme.svg?branch=master)](https://travis-ci.com/rust-lang/measureme)
+# measureme [![Rust](https://github.com/rust-lang/measureme/actions/workflows/rust.yml/badge.svg)](https://github.com/rust-lang/measureme/actions/workflows/rust.yml)
 Support crate for rustc's self-profiling feature
 
 This crate is maintained by the Rust compiler team and in particular by the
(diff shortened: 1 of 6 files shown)


michaelwoerister:
Thank you, @rylev!

@bors r+

michaelwoerister:
@wesleywiser this might be interesting for https://blog.rust-lang.org/inside-rust/2020/02/25/intro-rustc-self-profile.html too.

michaelwoerister (approved):

PR #195, opened 2022-04-07
Allow to build measureme on stable rust

marmeladema:
nightly rust was previously required because of the usage of the
unstable `asm` feature which was stabilized in version 1.59.0


wesleywiser (approved):
Awesome, thank you!

PR #200, opened 2022-07-26
Move profile data analysis into analyzeme from summarizeme

rylev:
This moves analysis of self profile data from the `summarize` crate to the `analyzeme` crate. This is needed for rustc-perf which will move from storing self profile data in the database to analyzing the data on demand. The changes can be summarized as follows:

* `summarize::analysis::perform_analysis` becomes `analyzeme::ProfileData::perform_analysis`
* `summarize::query_data::{Results, QueryData}` become `analyzeme::{AnalysisResults, QueryData}`
* Some code is shuffled in `summarize::diff` caused by the fact that the diff types and their non-diff counterparts no longer live in the same crate
* The diff types are moved into the diff module 
* This unfortunately makes `analyzeme` depend on `serde`. We could remove this dependency but we'd basically have to repeat `AnalysisResults` and `QueryData` definitions in summarizeme`.


Mark-Simulacrum (approved):
Overall looks good.

michaelwoerister (approved):

michaelwoerister:
Looks good to me. Since `summarize` in an executable (and thus no other crates can depend on it), we should be able to move stuff out of it without a major version bump, right?

rylev:
> Looks good to me. Since `summarize` in an executable (and thus no other crates can depend on it), we should be able to move stuff out of it without a major version bump, right?

That was my understanding, yes.

PR #205, opened 2023-03-20
Remove malformed `serde` attribute

dmah42:
```
error: unexpected end of input, unexpected token in nested attribute, expected ident
  --> crox/src/main.rs:35:13
   |
35 |     #[serde()]
   |     
```


michaelwoerister:
Thanks for the PR, @dmah42!

michaelwoerister:
The build failure on beta looks like https://github.com/rust-lang/rust/issues/109424. This should be fixed on nightly already.

PR #208, opened 2023-05-26
Allow whitespace control chars in EventId texts.

### LINKED ISSUE #207: Could not parse `event_id`. Found ASCII control character in <text>
I am not sure whether this is a rustc bug or a measureme bug so I'm reporting it here under the theory that it's more likely to be seen by the right people :)

I ran
```
x clean rustc_middle && MAGIC_EXTRA_RUSTFLAGS='-Zself-profile -Zself-profile-events=default,query-keys' x check --stage 0 rustc_middle
```
on https://github.com/rust-lang/rust/commit/ba6f5e3b4d60ea5a847cd4402cca594cd40b218f. That generated a .mm_profdata file that the measureme tools fail to parse:
```
; crox rustc_middle-2790234.mm_profdata
Could not parse `event_id`. Found ASCII control character in <text> at 112 in "erase_regions_ty^^impl for<'a, 'b> FnOnce(&'b mut DiagnosticBuilder<'a, ()>)
-> &'b mut DiagnosticBuilder<'a, ()>"
```
It looks like `FnOnce` impls are getting an extra newline generated at the end. That seems like a bug we should fix in rustc, but not such a bad bug that it should prevent measureme from generating a report - could we translated it to an ASCII space instead, maybe?

andjo403:
As it only seems that control chars is not allowed in the EventId texts is to allow for future tags I think excluding the whitespace control chars is ok.
What I can see right now the only tag used is for arguments and is '\x1E'.

fixes #207

cc @michaelwoerister

### DIFF
--- a/decodeme/src/event.rs
+++ b/decodeme/src/event.rs
@@ -93,7 +93,11 @@ impl<'a> Parser<'a> {
 
         self.pos = end;
 
-        if self.full_text[start..end].iter().any(u8::is_ascii_control) {
+        if self.full_text[start..end]
+            .iter()
+            .filter(|x| !x.is_ascii_whitespace())
+            .any(u8::is_ascii_control)
+        {
             return self.err("Found ASCII control character in <text>");
         }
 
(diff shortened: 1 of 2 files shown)


michaelwoerister:
Nice! Thanks, @andjo403 🙂

I'll need to take a closer look at the entire file format, to make sure this doesn't have any unintended consequences. It's been a while since I worked with this code. 

But I believe the suggested change matches the originally intended behavior.

andjo403:
did a test with assigning all the whitespace control chars to the label and args and crox seems to handle it correct at least.
`let event_label = profiler.alloc_string("l\ta\n\rbe\x0Cl");
let event_arg = profiler.alloc_string("a\tr\n\rg\x0Cs");`

image

but good that you look at the entire file format.

michaelwoerister:
OK, this looks good! Thanks again, @andjo403!

PR #221, opened 2024-01-10
Add GitHub Workflow for publishing measureme

wesleywiser:


michaelwoerister:
Thanks, Wesley!

@bors r+

.github/workflows/publish.yml
```diff
@@ -0,0 +1,17 @@
...
+on:
+  release:
+    types: [created]
+
+jobs:
+  publish:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
michaelwoerister, line 11: This looks like it might not do want we intend. IIUC, `actions/checkout@master` means that this will use the `master` branch version [actions/checkout](https://github.com/actions/checkout), not that it will checkout measureme's master branch. We probably want something like `actions/checkout@v4`, right? Ideally, we'd checkout the tag that was created for the release that triggered this.

michaelwoerister: > Ideally, we'd checkout the tag that was created for the release that triggered this.

OK, that seems to be the default behavior:
https://stackoverflow.com/questions/71663711/how-to-get-branch-name-which-triggered-a-github-action-workflow

Issue #229, opened 2024-05-28
Release analyzeme, decodeme, and measureme to crates.io


weihanglo:
During the course of [vendoring rustc-perf](https://github.com/rust-lang/rust/pull/125465) in rust-lang/rust. There were a couple of issues found:

* `analyzeme` and `decodeme` haven't yet published to crates.io.
* `analyzeme` depends on old versions of `measureme` via git branches. Those versions are not available on crates.io.

While it is possible to tweak rustc bootstrapping process to allow Git dependencies, it might be under the risk that a commit is missing or a branch points to a different commit.

## Proposed solution

The current https://github.com/rust-lang/rustc-perf/commit/892c6814e84785c43400093242ef6ed55ac72df9 depends on these packages:

```
└── analyzeme v11.0.0 (https://github.com/rust-lang/measureme?branch=stable#f2914503)
    ├── analyzeme v9.2.0 (https://github.com/rust-lang/measureme?tag=9.2.0#9f51cde2)
    │   ├── measureme v9.2.0 (https://github.com/rust-lang/measureme?tag=9.2.0#9f51cde2)
    ├── decodeme v10.1.2 (https://github.com/rust-lang/measureme?tag=10.1.2#f9f84d1a)
    │   ├── measureme v10.1.2 (https://github.com/rust-lang/measureme?tag=10.1.2#f9f84d1a)
    ├── decodeme v11.0.0 (https://github.com/rust-lang/measureme?branch=stable#f2914503)
    │   ├── measureme v11.0.0 (https://github.com/rust-lang/measureme?branch=stable#f2914503)
    ├── measureme v10.1.2 (https://github.com/rust-lang/measureme?tag=10.1.2#f9f84d1a) (*)
    └── measureme v11.0.0 (https://github.com/rust-lang/measureme?branch=stable#f2914503) (*)
```

I would suggest following these step and publishing theses packages if possible:

#### Version 9.2.1

1. Create a v9 branch from tag [9.2.0](https://github.com/rust-lang/measureme/releases/tag/9.2.0)
1. @weihanglo will create PR and changelog for the release https://github.com/rust-lang/measureme/compare/9.2.0...weihanglo:measureme:prepare-9.2.1
1. Review and merge the PR.
1. Publish `measureme@9.2.1` and `analyzeme@9.2.1` to crates.io (in this order).

#### Version 10.1.3

[...]

PR #230, opened 2024-05-29
refactor: inline packages for decoding v7 and v8 profdata files

weihanglo:
This is another step toward making rustc-perf self-contained.
See:

* https://github.com/rust-lang/measureme/issues/229#issuecomment-2136940676
* https://github.com/rust-lang/rust/pull/125465#issuecomment-2130745904

This is basically a slim version of vendoring older versions of those packages.
This is done by first vendoring, and then incrementally remove unused code.
Code is not shared between each version.


weihanglo:
@michaelwoerister this is review for review, though I am not sure if this is better than releasing old versions to crates.io 😅.

michaelwoerister:
> though I am not sure if this is better than releasing old versions to crates.io 😅.

Yes indeed. Looking at the size of the diff, I'm not so sure anymore either. I had overlooked that we also need to make a copy of the whole StringTable code. Frankly, I'd prefer if we kept things as they are and just released the old versions to crates.io. Backporting the license file changes is a one-time effort.

weihanglo:
Close as this is way more a headache than depending on old versions, and this approach is favored: <https://github.com/rust-lang/measureme/issues/229#issuecomment-2139227460>

PR #233, opened 2024-05-30
chore: prepare for v12 release

weihanglo:
Block on <https://github.com/rust-lang/measureme/pull/232> and <https://github.com/rust-lang/measureme/pull/234>.

Will update changelog after this PR is created.

PR #234, opened 2024-05-31
chore: publish also decodeme and analyzeme in CI

weihanglo:
This is a preparation of v12 release.

See #229 and #233


michaelwoerister (approved):
Thanks, @weihanglo!

Issue #237, opened 2024-11-19
Failed to build target `x86_64-unknown-linux-ohos` for crate


LuuuXXX:
Recently, I tried to build target `x86_64-unknown-linux-ohos`, and conditionally compiled `perf-event-open-sys` in the dependency. 
![Image](https://github.com/user-attachments/assets/e32df778-5182-4f9d-b9e9-8f92d5543662)

Here's an error report.
![Image](https://github.com/user-attachments/assets/09103bca-6f4f-4b3f-ad84-b36b14995207)

 I'd like to consult if this dependency is necessary, it seems to have been archived.
.

saethlin:
Per the diagnostic, this is a bug in `perf-event-open-sys`, or perhaps something even farther upstream than that.

LuuuXXX:
> Per the diagnostic, this is a bug in perf-event-open-sys

It seems perf-event-open-sys has been archived, perf-event hasn't been updated for a long time, may I replace it with perf-event-open-sys2.

PR #238, opened 2024-11-21
Cofigurate out `ohos` target's dependecies  to avoid compilation crashes

LuuuXXX:
I was trying to build ohos rustc targets. but blocked. It seems that` perf-event-open-sys` support is not good(https://github.com/rust-lang/measureme/issues/237#issuecomment-2484574815).
The new `perf-event-open-sys2` supports this which is an actively maintained fork.

cc https://github.com/rust-lang/measureme/issues/237

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ measureme_10 = { version = "10.1.3", package = "measureme" }
 memchr = "2"
 memmap2 = "0.2.1"
 parking_lot = "0.12.0"
-perf-event-open-sys = "3.0.0"
+perf-event-open-sys2 = "5.0.6"
 prettytable-rs = "0.10"
 rustc-hash = "1.0.1"
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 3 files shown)


LuuuXXX:
@saethlin Is it convenient to review the PR?

LuuuXXX:
@wesleywiser Is it convenient to review the PR?

Mark-Simulacrum:
Do these targets support perf counters? It looks like they're Linux systems and there's at least one x86_64 variant, but maybe we can just cfg out the dependency on this platform for now? That would unblock landing this without needing to consider a change in dependencies.

LuuuXXX:
thanks for review :)

>Do these targets support perf counters? 

yes supported.

>but maybe we can just cfg out the dependency on this platform for now? That would unblock landing this without needing to consider a change in dependencies.

ok, it's also make sense.

LuuuXXX:
I want to move towards the [MCP](https://github.com/rust-lang/compiler-team/issues/811) goal. I wonder if I can continue now. 
Will there publish a new version to matche this?

Mark-Simulacrum (approved):

LuuuXXX:
Will there be a new version released?

PR #246, opened 2025-07-07
Check `master` branch instead of `stable` in CI

Kobzol:
I don't think that the `stable` branch is useful anymore, now that we use actual crates.io releases and even the compiler depends on a normal crates.io version.


bjorn3:
Do we even need this job at all. If the format rustc uses changes, that necessarily means that this repo changed too.

oli-obk (approved):

PR #251, opened 2025-07-08
ci: add abililty to publish a single crate

marcoieni:


.github/workflows/publish.yml
```diff
@@ -22,5 +27,9 @@ jobs:
       - name: Publish to crates.io
         env:
           CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
-        run:
-          cargo +nightly publish -Zpackage-workspace --workspace
+        run: |
+          if [ -n "${{ github.event.inputs.crate }}" ]; then
+            cargo publish -p ${{ github.event.inputs.crate }}
```
marcoieni, line 32: I'm not using the nightly toolchain here. Let me know if you prefer to use it

Kobzol: You are using it, because it's the default in the workflow xD Let's use the same one in both cases.

Kobzol (approved):

